Issue #112, opened 2020-01-18
scaling up profileApply


brownag:
As has long been known, profileApply don't scale well.

Below is a first stab at speeding it up; benchmark times are on an old MacBook. 

Splitting the operation up into several runs of profileApply (resulting in more, smaller lists) seems to enhance performance significantly in the examples below with 1,000 and 10,000 random profiles. 
The chunk size is hardcoded at 100 profiles -- which seems to enhance performance in the thousands to tens of thousands. There is, as expected, some slight overhead relative to straight profileApply when using 100 profiles. Not sure what the upper end of this will be, but presumably as the number of chunks starts to get large, the effective rate of number of profiles applied will decrease.

This is run in a single thread, but this layout lends itself reasonably well to adding parallel computation option.

```
chunkApply <- function(object, FUN, simplify = FALSE, ...) {
  n <- length(object)
  grp <- sort(1:n %% max(1, round(n / 100))) + 1 
  res <- do.call('c', lapply(split(1:length(grp), grp), function(idx) {
    profileApply(object[idx, ], FUN, simplify = simplify, ...)
  }))
  names(res) <- profile_id(object)
  return(res)
}

library(aqp)

# generate 10,000 random profiles and promote to SPC
foo <- do.call('rbind', lapply(as.list(1:10000), random_profile))
depths(foo) <- id ~ top + bottom

# a "simple" function that returns a "complex" result
simpleFunction <- function(p) data.frame(horizons(p)[2,2:3])

c1 <- system.time(chunkApply(foo[1:100,], simpleFunction))
p1 <- system.time(profileApply(foo[1:100,], simpleFunction, simplify = FALSE))
c2 <- system.time(chunkApply(foo[1:1000,], simpleFunction))
p2 <- system.time(profileApply(foo[1:1000,], simpleFunction, simplify = FALSE))
c3 <- system.time(chunkApply(foo, simpleFunction))
p3 <- system.time(profileApply(foo, simpleFunction, simplify = FALSE))
```

100 PROFILES
```
c1
```
>   user  system elapsed 
>    0.254   0.000   0.267 
```
p1
```
>   user  system elapsed 
>    0.241   0.001   0.255 

1,000 PROFILES
```
c2
```
>  user  system elapsed 
>   2.335   0.028   2.689 
```
p2
```
> user  system elapsed 
>  5.057   0.024   5.596 

10,000 PROFILES
```
c3
```
> user  system elapsed 
>  29.734   0.236  33.839 
```
[...]

Issue #111, opened 2020-01-14
add 'frameify' argument to profileApply


brownag:
**Purpose:** facilitate handling of complex `profileApply` results by cleanly converting to data.frame, with _idname_ as column. Provides basic functionality for processing a list, with each list element containing a numeric vector with the values of several attributes -- essentially a single "row" or "record" in a data frame. 

In this example the vector contains the calculated top and bottom depth of PSCS for each pedon.

```
frameify <- function(l, colnames=NULL) {
  res <- as.data.frame(cbind(names(l), do.call('rbind', l)), row.names = FALSE)
  colnames(res) <- colnames
  return(res)
}

site(f) <- frameify(profileApply(f, estimatePSCS, simplify=FALSE), 
                    colnames = c(idname(f), "calc_pscstop", "calc_pscsbot"))
```

The above could be even further simplified by building the `frameify` result option into profileApply. Setting `frameify=TRUE` would override the default of `simplify=TRUE`. 

```
site(f) <- profileApply(f, estimatePSCS, frameify=TRUE, 
                    colnames = c(idname(f), "calc_pscstop", "calc_pscsbot"))
```

With not too much extra accounting this could be further extended to determine whether a site _or horizon_ level attribute is returned. 

For instance, if a hypothetical function returned a _matrix_ of values -- with one row/record for each pedon HORIZON --  `profileApply` could return a result that merges cleanly into the horizons slot by supplying both _idname_ and _hzidname_.

```
horizons(f) <- profileApply(f, somePedotransferFunction, frameify=TRUE, 
                    colnames = c(idname(f), hzidname(f), "newhzlevelattribute1", "newhzlevelattribute2"))
```

dylanbeaudette:
I like this. Could we also enforce "safe" merging of site / hz attributes by checking to see if one of the appropriate IDs are in the results? In other words, all functions provided to profileApply would have to be specific to an ID.

dylanbeaudette:
Pending a couple more tests, this is just about ready to go. Testing on 10k profiles:
 * 60 seconds using a single chunk
 * 14 seconds using chunk.size=100
 * 12 seconds using chunk.size=100 and pre-allocation of temporary lists

brownag:
Thanks for catch with stringsAsFactors being true by default. b4c171f -- issue closed

PR #113, opened 2020-01-19
upgrades to profileApply

brownag:
Addressing issues #111 #112 

Before upgrade:
  
```
> devtools::load_all()
Loading aqp
This is aqp 1.18.4
> foo <- do.call('rbind', lapply(as.list(1:2000), random_profile))
> depths(foo) <- id ~ top + bottom
> system.time(profileApply(foo, simpleFunction))
user  system elapsed 
18.893   0.160  19.967 
```

After "chunkApply" upgrade, nearly 4x faster processing of 2000 pedons.
  
```
> devtools::load_all()
Loading aqp
This is aqp 1.18.5
> foo <- do.call('rbind', lapply(as.list(1:2000), random_profile))
> depths(foo) <- id ~ top + bottom
> system.time(profileApply(foo, simpleFunction))
user  system elapsed 
5.528   0.057   5.837 
```

With `frameify = TRUE`:
  
```
> system.time(foo <- profileApply(foo, simpleFunction, frameify = TRUE))
user  system elapsed 
5.690   0.088   6.073 
```

"Simple" function that calculates horizon thickness and proportional thickness. 

Internal management of `idname` and `hzidname` by `frameify` facilitates taking stock of potentially incomplete results before _safely_ merging (e.g. using `site<-` or `horizons<-`) -- presuming `hzidname` is returned in the result from calling `FUN`.

``` 
simpleFunction <- function(p) {
  hz <- horizons(p)
  res <- data.frame(profile_id(p), hzID(p), thk=(hz$bottom - hz$top))
  res$hz_prop <- res$thk / sum(res$thk)
  # optional: return just second and third horizon to test merge result
  #res <- res[2:3,]
  colnames(res) <- c(idname(p), hzidname(p), 'hz_thickness', 'hz_prop')
  return(res)
}
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aqp
-Version: 1.18.4
-Date: 2020-01-06
+Version: 1.18.5
+Date: 2020-01-19
 Title: Algorithms for Quantitative Pedology
 Authors@R: c(person(given="Dylan", family="Beaudette", role = c("aut", "cre"), email = "dylan.beaudette@usda.gov"), person(given="Pierre", family="Roudier", email="roudierp@landcareresearch.co.nz", role = c("aut", "ctb")), person(given="Andrew", family="Brown", email="andrew.g.brown@usda.gov", role = c("aut", "ctb")))
 Author: Dylan Beaudette [aut, cre], Pierre Roudier [aut, ctb], Andrew Brown [aut, ctb]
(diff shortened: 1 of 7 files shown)


dylanbeaudette:
Nice work, this solves a long-standing bottleneck with all functions that rely on `profileApply`. Thanks.

Issue #118, opened 2020-02-07
proof of concept: magrittr %>% with SPCs


brownag:
# %>% %>%  aqp is "smokin'" %>% %>% 

[These](https://github.com/ncss-tech/aqp/commit/dcd18793ec15e488685cd3c2cd4823889fc0e0b3#diff-60cd57b48679b0d92578d1a70f52c139) (dcd18793e) are some prelim. results from my recent thought experiment on magrittr "pipelines" for SoilProfileCollections. 

## Overview

I am tentatively introducing new verbs exclusively for SPCs but not necessarily needing to be used with pipes. Or having these names. 

- logical comparison - ~subSPC()~ `filter` AND `subset` (preferred)
- NEW: transformation - `mutate`
- NEW: transformation by profile - `mutate_profile`
- NEW: calculate summaries - `summarize` 
- regular expression - `grepSPC`
- NEW: horizon depth interval based filtering - `glomApply`
- logical functional filtering - `subApply`

These functions operate on, and return, SoilProfileCollections. The tithe for this fancy feature is a new import for aqp, rlang, for tidy evaluation.

## [[ and [[<-

In this process, I found I also would need to implement a `[[` and `[[<-` method for the SPC. Both have proven to be quite handy in the implementation of these new filtering/transformation methods. These new S4 methods with double brackets are "ambivalent" and direct interfaces to site or horizon level attributes in the SPC. _Another_ benefit is that the `[[` interfaces can be iterated over by column name (in site and horizons)  , whereas `$`/`$<-` cannot and `[` is reserved for other purposes.

```
# identical $ and [[
loafercreek$clay == loafercreek[['clay']]

# convert clay to clay proportion with [[<-
loafercreek[['clay']] <- loafercreek$clay / 100
```

I'm open to comments/feedback/criticism :o: 

You will need rlang as a dependency, and magrittr if you want to use pipes.

---

Here's an example. 

```{r}
# load required packages
library(aqp)
library(soilDB)
library(magrittr)

# load sierra nevada foothills soils data
data("loafercreek")
data("gopheridge")

# create loafergopher dataset with aqp::union
loafergopher <- aqp::union(list(loafercreek, gopheridge))
```

## filter

```
# demonstrate horizon level subsetting with %>%
#  one or more horizons clay >35% and one or more frags >35%

g <- loafergopher %>%
  filter(clay >= 35 & total_frags_pct >= 35)

# filter can evaluate arbitrary # of logical expressions separated by comma
```

[...]

PR #120, opened 2020-02-08
SPC methods for use with magrittr %>% 

brownag:
# %>% %>%  aqp is "smokin'" %>% %>% 

PR for issue #118

[These](https://github.com/ncss-tech/aqp/commit/dcd18793ec15e488685cd3c2cd4823889fc0e0b3#diff-60cd57b48679b0d92578d1a70f52c139) (dcd18793e) are some prelim. results from my recent thought experiment on magrittr "pipelines" for SoilProfileCollections. 

## Overview

I am introducing three subsetting "verbs" exclusively for SPCs. They need not be used with pipes (%>%) but they are actually quite nice with them

-     regular expression -  `grepSPC()`

-     logical comparison -  `subSPC()`
 
-     functional evaluation - `subApply()`

These functions operate on, and return, SoilProfileCollections. The tithe for this fancy feature is a new import for aqp, rlang, for tidy evaluation.

## [[ subsetter

In this process, I found I also would need to implement a "[[" subsetter method for the SPC. It has proven to be quite handy. It's an "ambivalent" and direct interface to site or horizon level attributes in the SPC. _Another_ benefit is that for SPCs [[ can be iterated over by column name (in site and horizons)  , whereas $ cannot and [ is reserved for other purposes.

Did not need nor want to implement a setter at this time... but a [[-setter would have some clear advantages over `$<-`. Same general pattern as holds for the getters.
. 
```
# identical
loafercreek$clay == loafercreek[['clay']]
```

Branch "pipez" is working as intended, and passes R CMD CHECK without warning/notes.

```{r}
remotes::install_github("ncss-tech/aqp@pipez")
```

In the meantime, I'm open to comments/feedback/criticism :o: 

You will need: magrittr for `%>%` and rlang for tidy eval

---

Here's an example. 

```{r}
# load required packages
library(aqp)
library(soilDB)
library(magrittr)

# load sierra nevada foothills soils data
data("loafercreek")
data("gopheridge")

# create loafergopher dataset with aqp::union
loafergopher <- aqp::union(list(loafercreek, gopheridge))

# create taxon categorical variable for plotting
loafergopher$taxon <- factor(str_to_title(loafergopher$taxonname))
```

## subSPC

```
# demonstrate horizon level subsetting with %>%
#  one or more horizons clay >35% and one or more frags >35%

g <- loafergopher %>%
  subSPC(clay >= 35 & total_frags_pct >= 35)

# subSPC can evaluate arbitrary # of logical expressions separated by comma
```

It is easy to mix horizon and site level subsets in subSPC -- just put them in in separate expressions (comma-separated)

Can use any of the 3 functions together. They all just subset SoilProfileCollections.

```
f <- loafergopher %>%
      grepSPC(taxonname, "^Loaf|^Goph") %>%
      subSPC(taxonkind == 'taxadjunct', 
             taxorder == "alfisols") %>%
      subApply(is_clayey_sk)
```

## subApply

 subApply is a wrapper around profileApply. Argument is a function that takes a single profile SPC and returns a unit length logical. 

`is_clayey_sk` (defined below) returns a logical value representing whether or not a single profile is clayey skeletal (clay+frags both >35%) -- as shown in next block.

```
# define a detailed filtering function
# returns true if calculated PSCS clay+frags are >35%
is_clayey_sk <- function(p) {

  # estimate particle size control section
  pscs <- estimatePSCS(p)

  # glom horizons intersecting PSCS
  hz <- horizons(glom(p, z1 = pscs[1], z2 = pscs[2]))

  # truncate ragged edges to PSCS
  hz[hz$hzdept < pscs[1], 'hzdept'] <- pscs[1]
  hz[hz$hzdepb > pscs[2], 'hzdepb'] <- pscs[2] 

  # calculate weights
  hz$w <- hz$hzdepb - hz$hzdept

  # calculate weighted averages and test
  clay <- weighted.mean(hz$clay, w = hz$w) > 35
  frag <- weighted.mean(hz$total_frags_vol, w = hz$w) > 35

  return(clay & frag)
}
```

## grepSPC

The grepSPC function is fairly self explanatory; it provides access to pattern matching for any text-based field in site or horizons tables.

Overall what i am thinking is: if the type of filtering you need cant be reduced to a simple logical expression,  i.e. where you would use subSPC, then you can use subApply to bring that logic into the `%>%`-line. grepSPC provides for regex pattern matching.


dylanbeaudette:
This is great work. I really like the use of logical vectors for subsetting, but what happens when an expression is evaluated as `NA`? Do these functions automatically convert `NA` into `FALSE`?

brownag:
> This is great work. I really like the use of logical vectors for subsetting, but what happens when an expression is evaluated as `NA`? Do these functions automatically convert `NA` into `FALSE`?

Yes, the functions implicitly convert `NA` to `FALSE`. For site-level attributes this is done via `which()`, for horizon level attributes, the logical expressions are evaluated and matched to create a site level index, which any `NA` values are omitted from via `na.omit()`.

Issue #125, opened 2020-03-16
SPC j-index subsetting with valid spatial data


brownag:
I think the following logic (lines 857-881 in `SoilProfileCollection-methods.R`) should be removed.

There is nothing wrong with returning a SPC with only one horizon as a result from square-bracket subsetting. In fact, this happens occasionally when `glom`-ing over a programatically determined interval -- or at a single depth. 

This type of thing works fine with a SPC with uninitialized `@sp` slot but causes problems when you have valid spatial data.

Was there a particular reason why we couldn't return an SPC in this case? I know my use of the `j` index as an actual horizon index for `glom()` is a bit different from the original _sliced_-SPC model. 

Removing the code below would break some tests -- but I think it is probably quite rare that the automagic conversion to SPDF is used _intentionally_ in the wild.

```
    # if there is REAL data in @sp, and we only have 1 row of hz per coordinate- return SPDF
    # valid spatial data is now tested via validSpatialData()
    # also need to test that there is only 1 horizon/slice per location
    # only produces a SPDF when j index is present
    if(validSpatialData(x) & length(p.ids) == nrow(h) & !missing(j)) {
      # combine with coordinates
      message('result is a SpatialPointsDataFrame object')
      # note that we are filtering based on 'i' - an index of selected profiles
			
      # since the order of our slices and coordinates are the same
      # it is safe to use 'match.ID=FALSE'
      # this gets around a potential problem when dimnames(x)[[1]] aren't consecutive 
      # values-- often the case when subsetting has been performed
      
      ## TODO: there should always be something in @site
      # if site data, join hz+site
      if(nrow(s) > 0) {
      	return(SpatialPointsDataFrame(as(x, 'SpatialPoints')[i, ], data=join(h, s, by=idname(x)), match.ID=FALSE))
      }
      ## TODO: can this ever happen?
      # no site data
      else {
      	return(SpatialPointsDataFrame(as(x, 'SpatialPoints')[i, ], data=h, match.ID=FALSE))	
      }
    }
```

brownag:
For instance:

```
library(aqp)
library(soilDB)
data(loafercreek)

g1 <- glomApply(loafercreek, estimatePSCS)
length(g1) # works as "expected"

coordinates(loafercreek) <- ~ x_std + y_std
proj4string(loafercreek) <- "+proj=longlat +datum=WGS84"

g2 <- glomApply(loafercreek, estimatePSCS)
[...]

PR #127, opened 2020-03-16
remove implicit conversion to SPDF with unit length j-index #125

brownag:
PR to remove inconsistency in behavior of j-index subsetting with index of unit length for Spatial and non-Spatial SPCs

2nd commit: exploits the new stable behavior of single-horizon SPC results from j-index subsetting to implement a "modality" argument to manipulate `glom()` result


dylanbeaudette:
OK, in principle I like this idea: converting to a different class object should always be *explicit*. The original idea was perhaps more clever than its own good. I'm fine with this as long as we spend a little more time checking for potential bugs. I doubt anyone actually used this feature. I'll need a little more time to review the SPC tutorial and manual pages.

brownag:
Is there anything I can do to move this PR along?

Issue #135, opened 2020-05-25
optimizing various SoilProfileCollection methods


brownag:
```
library(profvis)
library(aqp)
library(soilDB)

data("loafercreek")

loafercreek$foo <- rep(1, nrow(loafercreek))

# make some test data -- this function is defined in @mollic branch
newdat <- permute_profile(loafercreek[1], n = 10000, boundary.attr = 'foo', min.thickness = 1)
```

brownag:
# [

```
test.fun <- function(n) {
  idx <- round(runif(n, 0, length(newdat)))
  return(newdat[idx,])
}

microbenchmark::microbenchmark(test.fun(10),
                               test.fun(100),
                               test.fun(1000),
                               test.fun(10000))
```

```
Unit: milliseconds
            expr      min       lq     mean   median       uq      max neval
    test.fun(10) 142.3052 147.5753 180.3458 164.2151 175.1167 557.4064   100
   test.fun(100) 146.5336 150.3979 179.9773 166.7710 172.6415 541.9834   100
  test.fun(1000) 169.5357 173.5430 203.8030 190.9973 195.3322 561.2541   100
 test.fun(10000) 263.1484 288.2051 319.6335 300.3561 316.3072 735.9368   100
```

brownag:
![image](https://user-images.githubusercontent.com/20842828/82769003-45117680-9de7-11ea-8e41-ca4cd9195c66.png)

brownag:
Here we see notable improvements (relative to above) from optimizing various parts of heavily used methods following above 2 commits (on @mollic branch) -- just by being careful to re-use variables and not invoke unnecessary calls to `unique` [if it can be avoided]. 
```
Unit: milliseconds
            expr       min        lq      mean    median        uq      max neval
    test.fun(10)  14.04452  15.02876  31.91002  18.14201  37.22526 356.3600   100
   test.fun(100)  16.70464  18.60879  52.88639  38.11054  44.03619 445.7565   100
  test.fun(1000)  35.08107  40.16366  96.29470  58.46718  75.18114 638.2684   100
 test.fun(10000) 116.34838 156.54107 219.95638 180.80176 231.97105 639.7117   100
```
[...]

Issue #89, opened 2019-03-15
square bracket indexing / subsetting should fail gracefully


dylanbeaudette:
Fix '[' method for `SoilProfileCollection` objects so that `x[, j]` does not fail or corrupt data when `x` has a profile with less than `j` horizons.

```r
library(aqp)

# simulate some data, IDs are 1:20
set.seed(1010101)
d <- lapply(1:20, random_profile)
d <- do.call('rbind', d )

# init SoilProfilecollection object
depths(d) <- id ~ top + bottom
head(horizons(d))

# differing numbers of horizons
profileApply(d, nrow)

plot(d)
plot(d[, 1:3])
plot(d[, 1:2])
plot(d[, 1:5])

# attempt to extract the nth horizon
plot(d[, 2])
plot(d[, 5])

## this works but the site data are wrong
profile_id(d[, 5])
site(d[, 5])
```

brownag:
The above condition is fixed in aqp/mollic. Propagates j-index constraints to site, diagnostics, restrictions and spatial slots.

Here is a demo:

```
devtools::install_github('ncss-tech/aqp@mollic')
library(aqp)
library(soilDB)

data("loafercreek")

loafercreek <- filter(loafercreek, !is.na(x_std))
coordinates(loafercreek) <- ~ x_std + y_std

# only a few profiles have n=7 or 8 horizons 
# none have 9 or 10
l7 <- loafercreek[,7:10]

par(mar=c(1,1,1,1))
plot(l7)

# here they are
plot(slot(l7, 'sp'))

# 39 sites
nrow(site(l7))

# 39 SpatialPoints
length(slot(l7, 'sp'))

# 127 diagnostic features
nrow(diagnostic_hz(l7))

# 39 unique profile ids in diagnostics
length(unique(diagnostic_hz(l7)[[idname(l7)]]))

# 49 restrictions
nrow(restrictions(l7))

# 39 unique profile ids in restrictions
length(unique(restrictions(l7)[[idname(l7)]]))
```
![image](https://user-images.githubusercontent.com/20842828/82792063-a7886800-9e23-11ea-94e9-c85e58f9f0b1.png)
![image](https://user-images.githubusercontent.com/20842828/82792175-d8689d00-9e23-11ea-8815-89daf6a4d077.png)

dylanbeaudette:
[...]

PR #136, opened 2020-05-27
Mollic

brownag:
**New functions and documentation pertaining to identification of diagnostic features and properties in SoilProfileCollections:**
- hasDarkColors, mollic.thickness.requirement
- depthOf. maxDepthOf, minDepthOf
- getCambicBounds
- guessHzAttrName, guessHzDesgnName, guessHzTexClName

**Experimental function for modifying horizon boundaries based on a "template" pedon and a vector of boundary distinctness standard deviations:**
- permute_profile

ALSO -- resolves long standing bug #89 and provides several enhancements to i and j indexing of SoilProfileCollections #135

Passing tests on R 3.6.3 (Windows + soilDB w/ local NASIS data )and R 4.0.0 (Ubuntu)


dylanbeaudette:
All tests pass on MacOS with R 3.6.1. 

Any reason for replacing `slot(x, 'site')` with `@` method? 

I'd like to look a little closer tonight, otherwise there is a huge pile of functionality in there, wow!

brownag:
RE: slot

"In the case of the slot function, name can be any expression that evaluates to a valid slot in the class definition. Generally, the only reason to use the functional form rather than the simpler operator is because the slot name has to be computed."

It is comparable in speed -- I will revert to using slot if you want ... I personally would prefer to use `@` internally wherever possible.

> In `estimateSoilDepth.R` I see that the error for bad horizon names has been removed and replaced with `guessHzDesgnName()`. What happens when `guessHzDesgnName()`fails to find a reasonable name and results in `NA` passed along? I like the approach, but there is no fail-safe now. I'd suggest a test for this condition and further checking / error in `estimateSoilDepth()`. Happy to help test this out.

Mmm that is indeed an oversight. Not sure why I wanted to allow it return `max(NULL)`. I fixed the guess functions so their output is a little more sensible in the pathological unguessable cases... but  less output is generated overall. There were a couple oversights in the flow of that function and `guessHzAttrName`.

> I like all of the new functions (e.g. `mollic.thickness.requirement`), but I do think that we should consider some naming conventions as we move forward. I know there is a mixture of dots and camelCase. I tend to prefer camelCase but don't really care. Lets discuss briefly.

I have generally been following the convention that my highest level functions use camel case, whereas lower level ones use dots. I.e. `getArgillicBounds` `crit.clay.argillic` `argillic.clay.increase.depth` `get.increase.matrix`. I suppose I just made that standard up from observing noise in the patterns of existing function naming... ha

> Looking over the test for `estimateSoilDepth()`, I don't recall that it returns `-Inf`. Am I reading this test correctly? I do think that there should be some kind of warning or error if the user specifies a bogus name to `estimateSoilDepth`. We should not automaticagically fix all input.

Test has been updated to reflect the error -- similar to that used to be generated when `name` was missing. 

```
library(aqp)
x <- data.frame(id=1, top=0:9, bottom=1:10, none=rep(NA, 10), one=rep(1, 10))
depths(x) <- id ~ top + bottom
plot(x)
estimateSoilDepth(x[1,])
```

```
> x <- data.frame(id=1, top=0:9, bottom=1:10, none=rep(NA, 10), one=rep(1, 10))
> depths(x) <- id ~ top + bottom
> plot(x)
unable to guess column containing horizon designations
> estimateSoilDepth(x[1,])
unable to guess column containing horizon designations
Error: soil depth estimation relies on a column containing horizon designations
```

ALSO, I would like to point out that I fixed S4 method dispatch for all SoilProfileCollection methods -- especially those that overlap with `dplyr` -- `aqp` should play nicer now if one is using both at same time -- regardless of which you load first.

dylanbeaudette:
Thanks for addressing those. I'm fine with using `@` for slot access, less typing. Function naming is fine with me although I've got a couple of functions that I'd like to rename eventually. Thanks for fixing S4 dispatch. I'll test all of this on my mac with the latest R today.

Issue #139, opened 2020-06-15
roxygen2 documentation for all S4 methods


brownag:
Many important internal methods are currently aliased to a single page --- and others are not explicitly documented other than in examples.

As I have worked to refine various structures of the dual site-horizon table model of the SoilProfileCollection object I will be implementing full documentation for the S4 methods that define the workings of the SPC.

This is part of a broader effort to migrate all documentation to roxygen (#28). It is hoped that the examples that follow demonstrate the value of that approach.

brownag:
Visual aid (example)

![image](https://user-images.githubusercontent.com/20842828/84670840-7ad3e780-aedb-11ea-9607-e1f8a0a1a2df.png)

dylanbeaudette:
Thanks. Any tips on converting long .Rd content into roxygen-style documentation? For example, the slab manual page.

brownag:
I have been doing 'em manually -- there might be a way to reverse engineer. I can pull together a default template for S4 methods. I have also worked out a general scheme for methods that have both a normal and a `<-` version.

I would be happy to do the slab/slice methods using the same method i have done so far. But so far have stuck to stuff previously aliased to SPC-utils or coercion.

brownag:
In general -- it is better to let Roxygen to do the heavy lifting. The main things that need to be lifted from the .Rd are the description/details, argument descriptions and examples. The rest is pretty boilerplate. In particular, autogenerated usage sections are far superior to hardcoded.

PR #142, opened 2020-06-17
Aqpdf

brownag:
add support for data.table and tibble


brownag:
OK. I literally cursed myself by mentioning in a github comment that i was way behind the master branch.

Only conflict is in description, which I will resolve. I'll run a few more tests this evening and I think the above commits are ready to go.

brownag:
> * in `compositeSPC` what is the purpose of converting factors to character?

Pattern matching and use of factor level names in logic rather than factors themselves. I suppose I did not discuss that, but it made sense for the functions where `compositeSPC` is used internally. The object that is composited is basically read-only, for what it is worth. It does not affect the SPC object itself.

> * lets talk about `gower_distance()`: this seems like a partial re-implementation of `profile_compare` without any of the benefits of the full implementation of Gower's distance (binary, nominal, ordinal distances) and the adjustments made by `profile_compare` to account for differences in depth.

Yikes, if i could have not even shown you that commit but i committed it. I will revert it. I do think that I considered all of those things you just mentioned -- not saying they are implemented in that code. 

> * I do not recall discussing `group_by` and the implicit manipulation of the object metadata

https://github.com/ncss-tech/aqp/issues/118#issuecomment-584491765

> * there are a lot of changes in SoilProfileCollecion-setters.R, I'd like to go over some of these. For example, in the replacement method for `horizons`, this kind of expression is really hard to follow: `sum(suppressWarnings(original.site.order != chnew)) > 0)`. I understand what it is doing, but implicit conversion of logicals to numeric is confusing. I know this is tiny, but `horizons` is fundamental to the entire package and thus extra care is required.

Would you prefer I write that in a different way? Or explicitly cast the type from logical to integer? I would be happy to discuss the logic. I am generally pleased with how `horizons` is working -- have you tried to break it yet? I had a couple scratchpad functions and am going to try and do a demoof some really cool functionality with `horizons<-`

brownag:
Sorry -- regarding `group_by` the object metadata -- that is at best a temporary implementation that could easily be replaced with a slot or attribute or some similar structure. It is only used with `summarize` and works well there. I do think we did discuss the use of the metadata for arbitrary purposes, which is why it is converted to list and that I am using it for ordering. So, we did talk about it. And I'm open to suggestions on how to implement it better.

dylanbeaudette:
Thanks for the replies. These answer most of my questions. Lets have a short conversation next time you are in the office. I'm fine with merging as long as you are confident in the testing.

brownag:
Please hold off on merging. I just wanted to make the pull request. It is much easier to track the progression of things in a PR. I suppose I will merge it when I am ready.

dylanbeaudette:
> Yikes, if i could have not even shown you that commit but i committed it. I will revert it. I do think that I considered all of those things you just mentioned -- not saying they are implemented in that code.

Nothing to be ashamed of, that is some very elegant code in there. Lets talk some more about it because I am about to do a total re-write of `profile_compare()`. Note that `profile_compare()` can accommodate site-level attributes.

brownag:
You mentioned that profile_compare needed review recently – and while I was working on various aspects of aqpdf I did spend some time looking at the comments and internals of it.  I wanted to understand what it really was doing internally… playing around and documenting some ideas and doing some comparisons against profile_compare. I like to try and implement things “for myself” as a way of developing understanding. It definitely had no place in this specific branch other than it was just a little aside I did late at night. Just was trying to see if I could write something that performs similarly without dependencies. I don’t want to replace or take over profile compare – and recognize I cannot hope to possibly reimplement all the great functionality provided by `cluster` -- but that doesn’t mean I can’t learn from trying things. 😊

I genuinely thought about the major issues associated with these types of distance metrics applied to soil profile collections – most/all of which was spurred on by your code, docs and comments. We def can talk about it – this is important stuff that needs to find its way into more common use. Based on your comments, there may be a bit more going on in mine than you realize – not saying the contents are either profound or obvious.

dylanbeaudette:
The algorithm is sound, tested, applied to large collections, and applied to a diverse sets of data. That said, there are a lot of inefficiencies in the code and some style-related issues that make me want to go back and give 2009-me a whack.

Most of the problems / TODO / ideas are outlined in #7 and the comments within profile_compare.R, as you found. I'm thinking about setting more default arguments (`max_d`, `k`) so that the function is simpler to use.

I'd like to discuss (later this month?) your thoughts on the matter and the novelties in `gower.R`. I spent some time last night looking over the weighting used, but as you say there could be something that I missed. If the two algorithms are sufficiently different, then I'd suggest giving it a name and we have both in aqp. There is no single numerical classification approach that solves every problem, so it is nice to have a couple options.

One small suggestion: add profile IDs and return as a `dist` object so that you can hang the profiles from a dendrogram. Or, arrange the profile IDs according to an ordination for an evaluation of proximity. Ordering sketches based on rows of the distance matrix is far less informative that would otherwise seem.

Another option, try `vegan::betadisper` for an interesting take on ordination.
image

R/SoilProfileCollection-integrity.R
```diff
@@ -139,7 +140,7 @@ setMethod('reorderHorizons',
                                    order(as.character(h[[idname(object)]]),
                                          h[[horizonDepths(object)[1]]]))
             
-            h <- aqp:::.as.data.frame.aqp(h[current.order,], 
+            h <- .as.data.frame.aqp(h[current.order,],
```
dylanbeaudette, line 143: Is this going to be necessary for all functions that modify `@horizons` or slot previously containing a `data.frame`?

man/profile_compare-methods.Rd
```diff
@@ -139,11 +139,20 @@ s$p2[1:na.required] <- NA
 f.check.NA <- function(i) length(which(is.na(i$p1) | is.na(i$p2))) / nrow(i) == 1
 missing.too.much.data.idx <- which(profileApply(s, f.check.NA))
 
-# remove bad profiles and try again: works
-s.no.na <- profile_compare(s[-missing.too.much.data.idx, ], vars=c('p1','p2'), 
-max_d=100, k=0, plot.depth.matrix=TRUE)
 
 
+# remove bad profiles and try again: works
```
dylanbeaudette, line 144: Nice catch. I'll review the current logic, wouldn't be surprised if it has missed something.

tests/testthat/test-plotSPC.R
```diff
@@ -169,4 +169,6 @@ test_that("addVolumeFraction fractional horizon depths", {
   
 })
 
-
+# cleanup
+if (file.exists("Rplots.pdf"))
```
dylanbeaudette, line 173: Thanks.

R/SoilProfileCollection-integrity.R
brownag: Yes, if you want to be sure that you preserve the datatype the user set. It incurs no unnecessary overhead unless something occurred that changed the type.

brownag:
Documentation (#139) is implemented for all of the fundamental internal S4 functions for the SPC class that manipulate slots etc. The R CMD check manual page problems have been fixed -- there were a couple mistakes I had made -- typos in names, missing things, duplications etc. 

I will probably need to write a Wiki page to make it easier / remember all the details for the future and make sure everyone knows how to do it.

I will be moving on to others (plots, slice/slab, profile_compare, etc) in the near future now that I have ironed out the process. roxygen is very straightforward with ordinary functions but the S4 methods require a lot more care. With S4, I think some of the things I am be doing are redundant relative to most modern version of roxygen and its ability to "guess" the right values -- so with some more testing I can probably pare down the markup further. 

Everything is passing except an example pertaining to removing missing data before using `profile_compare`. I commented it out just to be able to finish my check... but @dylanbeaudette can you please see if I am missing something obvious? Load my branch and try to run that part of the example. I think the check in `profile_compare` might be messed up. Link below...

Issue #92, opened 2019-03-21
normalize line endings


dylanbeaudette:
https://help.github.com/en/articles/dealing-with-line-endings

brownag:
Do we need to be more consistent about setting the relevant options in RStudio (or any other IDE) to minimize any unintentional changes to line endings? 

![image](https://user-images.githubusercontent.com/20842828/86624863-f2e07b00-bf78-11ea-9fbb-a33798288020.png)

This is perhaps part of the problem... but knowing which line ending to use is the bigger issue. Git can handle most of this if we set it up correctly. What is our "normalization" target? Using a .gitattributes file as you set up in AQP is the way to enforce this. Git to choosing via `text=auto`  might not be working out for us: possibly leads to the problems we have with all/most lines in files showing up as "changed" in a commit if the line ending type changes.

Currently we use `text=auto` but should we pick a specific target mode for source files?

We could try defaulting to `text eol=crlf` -- to enforce Windows-like behavior even when source is edited on OSX / Linux [or vice versa]. Then we could just say that when developing for aqp package then you should ensure all source files end with newline and use Windows line endings  (or whatever we decide is standard)

Are there any instances in the aqp package that specifically rely on a particular line ending? I could imagine it being an issue if we had specific build scripts for specific platforms or something like that -- but most if not everything should be platform ambivalent, right?

dylanbeaudette:
I'd prefer to standardize on LF EOL marker if at all possible. I suspect that Travis (linux-based) is barfing on the CRLF EOL marker.

brownag:
Yep -- I am leaning towards the same. Ill have it figured out -- I was catching some Rdas.

Looks like Travis doesnt really care either way about most things. Every error so far has been related to the binary files. There is a check in the build for ensuring LF in makefiles/batch scripts etc that need it, I believe.

brownag:
OK, with 7cba0c2 the binary files are no longer being corrupted when the repo is cloned fresh.

The CRLF -> LF conversion is implicit in the functioning of Git -- so it was redundant to use `text eol=lf` (as we decided on). Furthermore, the * was treating binary files to text even where explict rules were already present. I think the ordering was relevant -- and I had it in the wrong order originally. 

Now in .gitattributes we explicitly define a bunch of text/binary file extensions explicitly and do no use any general pattern that hits binary files too. This was a bit of a head scratcher for me -- and I didn't adequately test it in the case where the repo binary files were created anew.

[...]

PR #147, opened 2020-07-06
Renormalize

brownag:
This is an example of normalization that "chooses" the Windows-style CRLF line ending standard. Inverting it to chose LF would be simple.

- Sets .gitattributes to enforce CRLF line endings
- Renormalizes line endings for all affected text files

All unit tests pass and R CMD check is without error.

This would close issue #92 if the choice of standard is suitable. I think we do not need to be too prescriptive about IDE options as git checkout will handle any relevant conversions with ,gitattributes set -- but that may require some testing.


brownag:
So, the Travis CI build is not passing. Looks like there might need to be some files excluded from what I did above.

```

$ R CMD build  .
* checking for file ‘./DESCRIPTION’ ... OK
* preparing ‘aqp’:
* checking DESCRIPTION meta-information ... OK
* checking for LF line-endings in source and make files and shell scripts
* checking for empty or unneeded directories
* looking to see if a ‘data/datalist’ file should be added
Warning in readChar(con, 5L, useBytes = TRUE) :
  invalid or incomplete compressed data
Warning in readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
Warning in readChar(con, 5L, useBytes = TRUE) :
  invalid or incomplete compressed data
Warning in readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
Warning in readChar(con, 5L, useBytes = TRUE) :
  invalid or incomplete compressed data
Warning in readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
Warning in readChar(con, 5L, useBytes = TRUE) :
  invalid or incomplete compressed data
Warning in readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
Warning in readChar(con, 5L, useBytes = TRUE) :
  invalid or incomplete compressed data
Warning in readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
Error in if (any(update)) { : missing value where TRUE/FALSE needed
Execution halted
The command "R CMD build  ." failed and exited with 1 during .
```

brownag:
With a force-pushed to make renormalize branch match renormalize 2 from 9e49971 to c5f57aa this passes Travis CI checks. 

The original monkeying around I did hit the binary files too. Those commits are no longer a part of this PR after the force push. 

The following two commits contain the final fix
7b1d172
c5f57aa

Issue #79, opened 2019-01-25
reduce dependencies (if possible)


dylanbeaudette:
Relevant commentary:
   * http://dirk.eddelbuettel.com/blog/2018/02/28/#017_dependencies
   * http://www.tinyverse.org/ 

Some ideas on replacements:

   * `plyr` &rarr; `purrr` / `furrr`
   * `reshape` &rarr; `reshape2` (no longer maintained)
   * `stringr` &rarr; `stringi` (fewer dependencies)

Dropping `plyr` entirely is going to be very hard; `plyr::join` is used extensively.

Another approach, suggested by @smroecker: `plyr`, `reshape` &rarr; `data.table`

I am leaning towards the following:
   * `plyr::join` &rarr; `merge(..., sort=FALSE, incomparibles = NA)` (base) or equivalent from `data.table`
   * `plyr::ddply`  &rarr; `split` / `lapply` (I'll miss the syntactical sugar and NA-handling)
   * `melt` and `cast` / `dcast`  &rarr; [`data.table::dcast`](https://r-posts.com/benchmarking-cast-in-r-from-long-data-frame-to-wide-matrix/)
   * `plyr::rbind.fill` &rarr; `data.table::rbind.data.table(..., fill=TRUE)` 

**Update 2020-06-16**: `data.table` seems like the solution to most of our problems, and, gives us automatic parallel processing.

**Update 2021-01-20**: see #157 for `data.table`-specific changes and replacement of `reshape` + `plyr` functionality.

pierreroudier:
reshape → tidyr ?

dylanbeaudette:
Blurg... Haha, must resist hipsterverse.

dylanbeaudette:
Some ideas from [itdepends](https://speakerdeck.com/jimhester/it-depends)/

Still not my preferred syntax, but this felt rather natural:
```r
dep_usage_pkg('aqp') %>% 
count(pkg, sort=TRUE) %>% 
as.data.frame %>% 
knitr::kable(., row.names = FALSE)
```

```
|pkg          |    n|
|:------------|----:|
|base         | 5322|
|aqp          |  203|
|graphics     |  104|
|stats        |   62|
|grDevices    |   33|
|plyr         |   25|
|lattice      |   12|
|methods      |   11|
|cluster      |    8|
|reshape      |    7|
|scales       |    7|
|plotrix      |    5|
|sp           |    5|
|compositions |    4|
|grid         |    4|
|stringr      |    4|
[...]

PR #152, opened 2020-07-08
Integrity

brownag:
This PR resolves several known and unknown issues related to sorting based on profile ID and top depths by `depths<-`

As a result, tests that were changed from historic behavior e.g. for `sim` and the broken `tapply`-induced behavior of `profile_compare` are "fixed" [without altering the code of either of those methods].


brownag:
@dylanbeaudette see the .Rmd updated in c0c9fb4 for a summary of a whole bunch of the recent changes

brownag:
@dylanbeaudette see https://github.com/ncss-tech/aqp/pull/152/commits/1ff5a2c4612d049e826a28c1c7e551e4bbdfec0d#diff-e21ee3eb4c667b520ed503af4304d5b8

brownag:
The last several commits address several more issues and things that needed attending to.

Please merge these at your earliest convenience. They are stable and I believe generally an improvement over the master branch. There is a parallel change in soilDB package to merge that will fix .Rda files etc. for SPC changes.

 - convert `@hzdesgncol` and `@hztexclcol` S4 slots to metadata entries `aqp_hzdesgn` and `aqp_hztexcl` 
 - `aqp::union` should re-sort horizon/site and should use the default builder methods (dylan you were _right_ about this and I was a pompous ass about why it should be this way -- but union definitely should rebuild spc in a standard way -- whatever standard may be) 
 - figured out the issues with sp/rgdal warnings -- present on any rgdal 1.5-8 or higher -- suppressing the warnings in the S4 method. I don't think we are at risk of running afoul -- but my understanding of the matter is relatively limited: 
 - added mpspline2 demo [`spc.combined` demonstrates the ordering issue with _union'd_ SPCs, now fixed]

R/SoilProfileCollection-operators.R
```diff
@@ -122,27 +122,64 @@ setMethod("[", signature(x = "SoilProfileCollection",
...
+                  .idx.in.range <- function(len, jj) {
+                    1:len %in% jj
+                  }
+
+                  # this one was a head scratcher but is super cool! and pretty fast
+                  #  the main overhead comes from [,data.table ...
+                  # if(!data.table::is.data.table(h))
+                  #   h <- as.data.table(h)
+
+                  h <- h[, `:=`(.jmatch = .idx.in.range(.N, j),
```
dylanbeaudette, line 144: Whoa, this is new to me. I trust that this is well tested.

R/union.R
```diff
@@ -173,13 +198,21 @@ union <- function(spc=list(), method='all', na.rm=TRUE, drop.spatial=FALSE) {
...
 
   # generate new SPC components
-  # using plyr::rbind.fill seems to solve the problem on non-conformal DF
   # https://github.com/ncss-tech/aqp/issues/71
-  o.h <- do.call('rbind.fill', o.h) # horizon data
-  o.s <- do.call('rbind.fill', o.s) # site data
-  o.d <- do.call('rbind.fill', o.d) # diagnostic data, leave as-is
-  o.r <- do.call('rbind.fill', o.r) # restriction data, leave as-is
-
+  if(requireNamespace("data.table")) {
```
dylanbeaudette, line 202: Nice.

R/union.R
```diff
@@ -225,23 +258,25 @@ union <- function(spc=list(), method='all', na.rm=TRUE, drop.spatial=FALSE) {
...
+  # take combined horizon data, convert to target df subclass
+  res <- .as.data.frame.aqp(o.h, o.df.class)
+
+  ## reset horizon IDs (if the default column hzID is presnt)
+  if ('hzID' %in% colnames(res))
+    res$hzID <- NULL
+
+  # rebuild SPC using safe constructors
+  depths(res) <- formula(sprintf("%s ~ %s + %s", new.pID, new.hzd[1], new.hzd[2]))
+  site(res) <- .as.data.frame.aqp(o.s, o.df.class)
```
dylanbeaudette, line 270: This is the extra complexity that I am talking about: is this required every time `site(x) <- ` is invoked?

dylanbeaudette:
> The last several commits address several more issues and things that needed attending to.
> 
> Please merge these at your earliest convenience. They are stable and I believe generally an improvement over the master branch. There is a parallel change in soilDB package to merge that will fix .Rda files etc. for SPC changes.
> 
> * convert `@hzdesgncol` and `@hztexclcol` S4 slots to metadata entries `aqp_hzdesgn` and `aqp_hztexcl`

I'm curious why the change of direction here: more flexibility for future change?

> * `aqp::union` should re-sort horizon/site and should use the default builder methods (dylan you were _right_ about this and I was a pompous ass about why it should be this way -- but union definitely should rebuild spc in a standard way -- whatever standard may be)
> * figured out the issues with sp/rgdal warnings -- present on any rgdal 1.5-8 or higher -- suppressing the warnings in the S4 method. I don't think we are at risk of running afoul -- but my understanding of the matter is relatively limited: https://cran.r-project.org/web/packages/rgdal/vignettes/PROJ6_GDAL3.html
> * added mpspline2 demo [`spc.combined` demonstrates the ordering issue with _union'd_ SPCs, now fixed]

R/union.R
brownag: This is `union`... No.

brownag: Also, .as.data.frame.aqp only checks the class. we have talked about this before. there is no overhead or complexity to speak of,. you read my document that covered .as.data.frame.aqp and .data.frame.j. These are the two features that were added

brownag:
> * Good idea moving imports -> suggests, this has been on a TODO list for a long time. We may need to talk about digest: as there is a strong reason to keep this in imports, but a stronger reason to remove plyr. Fine for now.

As far as I could tell `diges`t was only used for `aqp::unique` -- which is not commonly used functionality. The package runs just fine without it in imports. And its a common package. So I'm curious what the strong reason is that it needs to be imported? We have a lot of imports for what appear to be not-that-important reasons... but maybe I am missing something?

> * Switching from slots -> metadata is a big move, I see how this can prevent future mishaps (invalid SPC structure) at the S4 level, but we should discuss a long-term plan. When to use slots / when to stuff things into the new list representation of `metadata`. 

I mean, it was an ill advised move to put them there. You probably should have suggested I put them in metadata in the first place... and I am pretty sure I didnt give you taht opportunity I just committed it to master. This is how it should be. I

I added those slots as optional -- it never made sense because they are optional. This process of slots -> metadata stops here.... Indeed I wonder whether some of my new actually IMPORTANT stuff should be slots (target.order, aqp_df_class, aqp_group_by, etc)

You are right that this smacks of S3. But we are not converting to S3, or at least I will not have a hand in it in this package. I will likely write a S3 representation for soil profile data soon just to show that it can be done and weigh the options more effectively. 

> * Another question regarding moving slot contents to `metadata`: is it still possible to replace `metadata` all at once, possibly corrupting the previously registered horizon designations etc.? Should we maintain two levels of mutability in SPC objects?

No. You were the only one to do that. It would be trivial to modify the metadata<- function or add a replacement method. You should use `metadata(object)$name <- value` ... I can hardly think of any reason to replace the whole thing if you are not an internal AQP method. It is a good idea though -- there are a couple places where old and new metadata get merged and those could be abstracted. I would encourage you to play around with this in your workflows and see if it works well enough and if more functionality is needed. Pretty sure it is just you and me who "use" metadata in code

> * regarding flexibility in slot contents (`data.frame`, `data.table`, `tibble`): is the complexity of supporting all three adding significant complexity to the code? I'd rather supporting just `data.table` for the performance gains vs. overly complex code.

There is not significant added complexity. Inheritance is a very beautiful thing. There are major benefits from having all three and a safe revert to base R. You will probably need to spend some time figuring out how to use data.table as I have had to correct a lot of code that used j index or single bracket column name subsetting ambiguously.

You have raised this and the above points a couple times saying they need to be talked about ... but that was several months ago. You should really spend some time considering ways to integrate`data.table` under the paradigm we discussed in the past -- carefully inserting it where we can [using `requireNamespace`] and then eventually possibly "pulling off the training wheels." 

I think its time we just start trying stuff rather than hypothesizing about what it might look or work like, I literally do not know what an aqp package that just supports data.table would look like -- please give me some pointers. I would think it would be more effort to exclude the possibility of data.frame, but thats just me.

I really do not think that aqp should depend on things like reshape, plyr or data.table -- though it should take advantage where it can. Every single key functionality should be achievable in base R. We are of course not there yet.

> * Did you run the code through some kind of linter? I see a lot of changes that look like re-formatting. I'm fine with that but would appreciated a heads-up. For example, I see a lot of lines hard-wrapped at 80 (?) columns. That is really narrow on most modern displays.

Negative. I have used RStudios re-flow code a couple times on individual functions. The changes I think you are referring to are removal of whitespace on empty lines... a result of .gitattributes changes I think after saving files. I hard wrap my lines manually. Usually around 90. It is a habit that I bet you don't like and I can try to stop doing it for my contributions to aqp.

> * is it really necessary to issue the message "unsorted input data will be ordered by profile ID and top depth" at SPC init time?

I guess not. I think that these types of things should be explicit though. This has been fundamental to the logic of SPC forever and I bet most users arent aware of it. Canning a message is easy though.

R/SoilProfileCollection-operators.R
brownag: I guess that means you didn't run the "big random" demo I tagged a couple days ago. Try it.

dylanbeaudette: I reviewed it, and running something doesn't confer understanding of the implementation. I still have not had time to review much of the `data.table` functionality, other  than what is related to the stuff I used in my re-write of `slice`. My comment had to do with the syntax: I have never seen `.idx.in.range` used in this context.

R/union.R
dylanbeaudette: I recall those conversations. To be clear: at SPC construction time (e.g. I take an SPC apart, do something, and put it back together) can I use standard constructors without making special accommodation for `data.table` or `tbl_df` in the general sense? My concern is this: added complexity at SCP construction / deconstruction is something that I want to avoid.

dylanbeaudette:
> > * Good idea moving imports -> suggests, this has been on a TODO list for a long time. We may need to talk about digest: as there is a strong reason to keep this in imports, but a stronger reason to remove plyr. Fine for now.
> 
> As far as I could tell `diges`t was only used for `aqp::unique` -- which is not commonly used functionality. The package runs just fine without it in imports. And its a common package. So I'm curious what the strong reason is that it needs to be imported? We have a lot of imports for what appear to be not-that-important reasons... but maybe I am missing something?

Please keep in mind that your usage of `aqp` is not necessarily representative of all uses. I was trying to be diplomatic in saying that swapping around imports / suggests is a fine thing to run by me. Look, I know that this is a long-standing TODO item, but that is a rather large change in terms of maintaining documentation and dependencies. It turns out that digest / unique is used pretty heavily in some of my work and in code supporting SoilWeb.

> > * Switching from slots -> metadata is a big move, I see how this can prevent future mishaps (invalid SPC structure) at the S4 level, but we should discuss a long-term plan. When to use slots / when to stuff things into the new list representation of `metadata`.
> 
> I mean, it was an ill advised move to put them there. You probably should have suggested I put them in metadata in the first place... and I am pretty sure I didnt give you taht opportunity I just committed it to master. This is how it should be. I
> 
> I added those slots as optional -- it never made sense because they are optional. This process of slots -> metadata stops here.... Indeed I wonder whether some of my new actually IMPORTANT stuff should be slots (target.order, aqp_df_class, aqp_group_by, etc)
> 
> You are right that this smacks of S3. But we are not converting to S3, or at least I will not have a hand in it in this package. I will likely write a S3 representation for soil profile data soon just to show that it can be done and weigh the options more effectively.

How about just mentioning it in the future. We should all be careful about breaking-changes to the SPC: it can cause all manner of downstream problems, esp. with CRAN. I appreciate the excitement and energy, but we need to use some caution. I agree that some things are more "optional" than others, esp. those components that are more relevant to internal NCSS stuff vs. the general case. I'll put some of my thoughts into a new issue dedicated to this discussion (https://github.com/ncss-tech/aqp/issues/154).

> > * Another question regarding moving slot contents to `metadata`: is it still possible to replace `metadata` all at once, possibly corrupting the previously registered horizon designations etc.? Should we maintain two levels of mutability in SPC objects?
> 
> EDIT: YES you can replace it but you shouldn't. You were the only one to do that. It would be trivial to modify the metadata<- function or add a replacement method. You should use `metadata(object)$name <- value` ... I can hardly think of any reason to replace the whole thing if you are not an internal AQP method. It is a good idea though -- there are a couple places where old and new metadata get merged and those could be abstracted. I would encourage you to play around with this in your workflows and see if it works well enough and if more functionality is needed. Pretty sure it is just you and me who "use" metadata in code

Please recall that the concept / implementation of `@metadata` has evolved considerably over 10 years. I'd suggest we formalize its purpose now. I'd rather see getter/setter methods for some of the more specific "settings" like horizon designation. The key/value concepts is very flexible `metadata(object)$name <- value`, but all it takes is a typo to clobber something important. Someone could have made the same mistake with slots, but it is a one step removed. I agree that full replacement of `@metadata` is dangerous and should be removed from the examples (I'll do that).

> > * regarding flexibility in slot contents (`data.frame`, `data.table`, `tibble`): is the complexity of supporting all three adding significant complexity to the code? I'd rather supporting just `data.table` for the performance gains vs. overly complex code.
> 
> There is not significant added complexity. Inheritance is a very beautiful thing. There are major benefits from having all three and a safe revert to base R. You will probably need to spend some time figuring out how to use data.table as I have had to correct a lot of code that used j index or single bracket column name subsetting ambiguously.

OK. I asked because I "see" class-wise adjustments in several places now. I'll return to this if I run into problems. For now, the "no it does not" answer is what I was looking for.

> You have raised this and the above points a couple times saying they need to be talked about ... but that was several months ago. You should really spend some time considering ways to integrate`data.table` under the paradigm we discussed in the past -- carefully inserting it where we can [using `requireNamespace`] and then eventually possibly "pulling off the training wheels."

Yes, it is different, and yes I have spent time with it. A good reminder to finish the re-write of `slice`. Keep in mind that my entire day cannot be filled with `aqp` development, as fun as that might be.

> I think its time we just start trying stuff rather than hypothesizing about what it might look or work like, I literally do not know what an aqp package that just supports data.table would look like -- please give me some pointers. I would think it would be more effort to exclude the possibility of data.frame, but thats just me.

Sure, I'm all for trying things, but I am opposed to flip-flopping internals from one extreme to another without some serious consideration. That is a slow process, sorry. My original vision was to make use of `data.table` as-needed, for the boost in speed indexing large collections, and for the reshaping functionality. I agree that forcing people to use `data.table` is heavy handed and perhaps opens the door to all manner of bugs when interfacing with base/external packages that expect pure `data.frame`s. At this point navigating the trade-offs is more important than the specifics--that will come next.

> I really do not think that aqp should depend on things like reshape, plyr or data.table -- though it should take advantage where it can. Every single key functionality should be achievable in base R. We are of course not there yet.

Agreed, that is what (https://github.com/ncss-tech/aqp/issues/79) is all about. I am not in favor of re-inventing the wheel when it comes to indexed sub-setting, parallel processing, and difficult reshaping. Therefore, I'd gladly trade `reshape` and `plyr` for `data.table`. Dependencies aren't evil but have to be carefully managed. I've slowly been removing all use of `plyr` and the time has come to finish that process and remove `reshape` as well. Attempting to do *most things* in base R is a noble goal that we can partially achieve. Attempting to do *everything* in base R is foolish, hence packages.

> > * Did you run the code through some kind of linter? I see a lot of changes that look like re-formatting. I'm fine with that but would appreciated a heads-up. For example, I see a lot of lines hard-wrapped at 80 (?) columns. That is really narrow on most modern displays.
> 
> Negative. I have used RStudios re-flow code a couple times on individual functions. The changes I think you are referring to are removal of whitespace on empty lines... a result of .gitattributes changes I think after saving files. I hard wrap my lines manually. Usually around 90. I do not use soft-wrap. It is a habit that I bet you don't like and I can try to stop doing it for my contributions to aqp.

OK. Thanks for the clarification. Coding styles are different and I respect that. I'll voice my main comment as "please don't run our code through a linter".

> > * is it really necessary to issue the message "unsorted input data will be ordered by profile ID and top depth" at SPC init time?
> 
> I guess not. I think that these types of things should be explicit though. This has been fundamental to the logic of SPC forever and I bet most users arent aware of it. Canning a message is easy though.

I agree, and see the utility. Can you think of a case where not seeing this message would lead to disaster? How about we just mention it in the manual page? Overly verbose functions can be very annoying.

brownag:
> > > * Good idea moving imports -> suggests, this has been on a TODO list for a long time. We may need to talk about digest: as there is a strong reason to keep this in imports, but a stronger reason to remove plyr. Fine for now.
> > 
> > 
> > As far as I could tell `diges`t was only used for `aqp::unique` -- which is not commonly used functionality. The package runs just fine without it in imports. And its a common package. So I'm curious what the strong reason is that it needs to be imported? We have a lot of imports for what appear to be not-that-important reasons... but maybe I am missing something?
> 
> Please keep in mind that your usage of `aqp` is not necessarily representative of all uses. I was trying to be diplomatic in saying that swapping around imports / suggests is a fine thing to run by me. Look, I know that this is a long-standing TODO item, but that is a rather large change in terms of maintaining documentation and dependencies. It turns out that digest / unique is used pretty heavily in some of my work and in code supporting SoilWeb.

Of course you should keep this in mind too. Why does SoilWeb have any bearings on what aqp imports, aside from additional library calls or specific function access you might need to make? I guess I am just struggling to understand your point here... also being diplomatic.

> > > * Switching from slots -> metadata is a big move, I see how this can prevent future mishaps (invalid SPC structure) at the S4 level, but we should discuss a long-term plan. When to use slots / when to stuff things into the new list representation of `metadata`.
> > 
> > 
> > I mean, it was an ill advised move to put them there. You probably should have suggested I put them in metadata in the first place... and I am pretty sure I didnt give you taht opportunity I just committed it to master. This is how it should be. I
> > I added those slots as optional -- it never made sense because they are optional. This process of slots -> metadata stops here.... Indeed I wonder whether some of my new actually IMPORTANT stuff should be slots (target.order, aqp_df_class, aqp_group_by, etc)
> > You are right that this smacks of S3. But we are not converting to S3, or at least I will not have a hand in it in this package. I will likely write a S3 representation for soil profile data soon just to show that it can be done and weigh the options more effectively.
> 
> How about just mentioning it in the future. We should all be careful about breaking-changes to the SPC: it can cause all manner of downstream problems, esp. with CRAN. I appreciate the excitement and energy, but we need to use some caution. I agree that some things are more "optional" than others, esp. those components that are more relevant to internal NCSS stuff vs. the general case. I'll put some of my thoughts into a new issue dedicated to this discussion (#154).

I did mention it to you dude. And this will not cause downstream problems aside from breaking SPC. Which I have done several times since last CRAN commit. I would like it to be sorted out properly for the next one!!

> > > * Another question regarding moving slot contents to `metadata`: is it still possible to replace `metadata` all at once, possibly corrupting the previously registered horizon designations etc.? Should we maintain two levels of mutability in SPC objects?
> > 
> > 
> > EDIT: YES you can replace it but you shouldn't. You were the only one to do that. It would be trivial to modify the metadata<- function or add a replacement method. You should use `metadata(object)$name <- value` ... I can hardly think of any reason to replace the whole thing if you are not an internal AQP method. It is a good idea though -- there are a couple places where old and new metadata get merged and those could be abstracted. I would encourage you to play around with this in your workflows and see if it works well enough and if more functionality is needed. Pretty sure it is just you and me who "use" metadata in code
> 
> Please recall that the concept / implementation of `@metadata` has evolved considerably over 10 years. I'd suggest we formalize its purpose now. I'd rather see getter/setter methods for some of the more specific "settings" like horizon designation. The key/value concepts is very flexible `metadata(object)$name <- value`, but all it takes is a typo to clobber something important. Someone could have made the same mistake with slots, but it is a one step removed. I agree that full replacement of `@metadata` is dangerous and should be removed from the examples (I'll do that).

I think I have corrected your examples already. And have provided accessor methods e.g. aqp_df_class. There are setters for e.g. depth_units.

> > You have raised this and the above points a couple times saying they need to be talked about ... but that was several months ago. You should really spend some time considering ways to integrate`data.table` under the paradigm we discussed in the past -- carefully inserting it where we can [using `requireNamespace`] and then eventually possibly "pulling off the training wheels."
> 
> Yes, it is different, and yes I have spent time with it. A good reminder to finish the re-write of `slice`. Keep in mind that my entire day cannot be filled with `aqp` development, as fun as that might be.

I am not pushing you to incorporate these changes. But just simply asking that you respect that I have considered these things on some level and my work is consistent with 

> > I think its time we just start trying stuff rather than hypothesizing about what it might look or work like, I literally do not know what an aqp package that just supports data.table would look like -- please give me some pointers. I would think it would be more effort to exclude the possibility of data.frame, but thats just me.

> > I really do not think that aqp should depend on things like reshape, plyr or data.table -- though it should take advantage where it can. Every single key functionality should be achievable in base R. We are of course not there yet.
> 
> Agreed, that is what (#79) is all about. I am not in favor of re-inventing the wheel when it comes to indexed sub-setting, parallel processing, and difficult reshaping. Therefore, I'd gladly trade `reshape` and `plyr` for `data.table`. Dependencies aren't evil but have to be carefully managed. I've slowly been removing all use of `plyr` and the time has come to finish that process and remove `reshape` as well. Attempting to do _most things_ in base R is a noble goal that we can partially achieve. Attempting to do _everything_ in base R is foolish, hence packages.

Sure. I think the important word is "key" functionality. doing md5 digests of Profiles is not really "key" functionality, for instance, whereas the single bracket subsetter is.

> I agree, and see the utility. Can you think of a case where not seeing this message would lead to disaster? How about we just mention it in the manual page? Overly verbose functions can be very annoying.

I just removed it. Agreed. The docs should be super clear about it though.

brownag:
Also, I really take issue with your comment about not having your full day available for aqp development. I put a lot of of my own time into this -- but I certainly don't expect you to have a full day available.  There is nothing about me submitting a PR that says you have to merge it before you have had time to fully review it. Just sayin'

Youve merged a bunch of my PRs in past couple months only to come back and question the contents later. Please consider that it is worth just taking the time when you can. If you need me to stop making changes to aqp, I can direct my efforts elsewhere. I am probably converging on the limits of the changes I will be making here now that this PR is merged

R/SoilProfileCollection-operators.R
brownag: .idx.in.range is an internal new function that is defined right above that call. So I don't know where you would have seen it other than in my code.

PR #155, opened 2020-07-14
Optimization of [,SoilProfileCollection-method and fixes for data.table implementation

brownag:
Some updates following recent merge of `integrity` branch;

 - gets rid of hack-y `:=` mask (no more messages)
 - defines _data.table_ special functions `.N` `.I` as needed for R CMD CHECK within local (not global) context
 - new i and j index determination (2-3x faster for 1M `random_profile` SPC) for _data.table_ SPCs that takes advantage of more appropriate _data.table_ features
 - trial method for enforcing DF subclasses and setting keys on _data.table_ SPCs
 - clarify logic
 - update bigrandom demo


R/enforce_df_class.R
```diff
@@ -0,0 +1,39 @@
...
+      stop("package `data.table is required", call. = FALSE)
+
+    # set keys
+    data.table::setkeyv(object@horizons, c(idname(object), horizonDepths(object)[1]), verbose = FALSE)
+    data.table::setkeyv(object@site, idname(object), verbose = FALSE)
+
+    if(length(colnames(object@diagnostic) > 1))
+      data.table::setkeyv(object@diagnostic, idname(object), verbose = FALSE)
+
+    if(length(colnames(object@diagnostic) > 1))
```
dylanbeaudette, line 34: copy / paste typo?

what happens slots contain 0-row `data.frames`?

misc/aqp2/big_random.R
```diff
@@ -1,62 +1,88 @@
...
-save(bigrandom, file="bigrandom.Rda")
+#
+# # build the data
+# system.time(bigrandom <- lapply(1:1000001, random_profile))
+# # user  system elapsed
+# # 933.54    1.95  939.94
+#
+# # save for later
+#
+load(file = "~/bigrandom.Rda")
```
dylanbeaudette, line 13: This is in /misc so not a big deal, but `save` / `load`? and indexing to $HOME?

R/enforce_df_class.R
brownag: Yes, if the restrictions and or diagnostics slots are empty data frames you get an error trying to set key

misc/aqp2/big_random.R
brownag: Yeah -- sorry -- not sure what you are asking there. This is just a byproduct of me commiting changes between my linux machine and work and not consistently fixing the paths. The latest version should just reference "bigrandom.Rda" unless I messed up.

R/enforce_df_class.R
brownag: I don't think there is any issue with zero row as long as the column exists -- which it generally should for all of our slots. There are no legitimate ways to add things to diagnostics or restrictions without having a matching idname in site.

brownag:
This branch has had its history rebased, FYI. Should have resulted in essentially the same working tree -- but some things were fixed, Some files you previously commented will show up as outdated now.

misc/aqp2/big_random.R
dylanbeaudette: Hah, ok. I had a couple of free minutes. I'll wait a day or two next time.

R/enforce_df_class.R
dylanbeaudette: OK, thanks for checking.

R/SoilProfileCollection-operators.R
```diff
@@ -85,14 +85,14 @@ setMethod("[", signature(x = "SoilProfileCollection",
...
-            h <- .as.data.frame.aqp(x@horizons, aqp_df_class(x))
+            h <- x@horizons
             s.all <- x@site
 
             # extract requested profile IDs
             p.ids <- s.all[[idname(x)]][unique(i)]
 
             # keep only the requested horizon data (filtered by profile ID)
-            h <- h[h[[idname(x)]] %in% p.ids,]
+            h <- .as.data.frame.aqp(h, aqp_df_class(x))[h[[idname(x)]] %in% p.ids,]
```
dylanbeaudette, line 95: Is this necessary every time the contents of `@horizons` is extracted / filtered based on row or column expressions?  I want to make sure that I understand the full implications of supporting 3 possible classes.

misc/aqp2/big_random.R
brownag: Its OK, i should have committed the changes I had on deck. I didn't think I had hit the review button, but was planning to.

R/SoilProfileCollection-methods.R
```diff
@@ -80,12 +80,41 @@ setMethod(
     }
 
     # filter out missing data
-    h <- .data.frame.j(h, target.names, aqp_df_class(x))
+    h <- aqp:::.data.frame.j(h, target.names, aqp_df_class(x))
```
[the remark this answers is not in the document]
brownag: You cannot use a vector of column names in the j index in the same way that you can in a data.frame. It might "work" but not the same way -- often it fails with a verbose error.

R/SoilProfileCollection-methods.R
```diff
@@ -80,12 +80,41 @@ setMethod(
...
+        # 7.26    5.52   16.83 
+        # # with pre-set key
+        # user  system elapsed 
+        # 2.07    0.00    2.08  
+        
+        # cant invoke this for something like min/max probably -- might do better on linux
+        # data.table::setkeyv(h, c(idn))
+        
+        dep <- h[[hz_bottom_depths]]
+        d <- dep[h[, .I[hz_bottom_depths == max(hz_bottom_depths, na.rm = T)][1],
```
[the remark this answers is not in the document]
brownag: I have no idea. The efficiency is about 2x better than tapply. I think I could possibly do it better with `:=` but I really struggled to improve performance on the million profile case. This was as good as it got and it didnt seem much better than tapply, even when the keys are set.

R/SoilProfileCollection-methods.R
```diff
@@ -80,12 +80,41 @@ setMethod(
...
+                     by = idn]$V1] 
+        
+        # return from here for data.table
+        return(min(d, na.rm = TRUE))
+      }
+    }
```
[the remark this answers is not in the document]
brownag: correct, because of the only modest improvement for data.table -- it is not outweighed by setting a key -- I don't think

It was something like 5-6 seconds for tapply, versus 2 seconds for data.table keyed join, versus like 16 seconds for setting the key + doing the join

R/SoilProfileCollection-operators.R
```diff
@@ -85,14 +85,14 @@ setMethod("[", signature(x = "SoilProfileCollection",
             }
 
             # extract all site and horizon data
-            h <- .as.data.frame.aqp(x@horizons, aqp_df_class(x))
+            h <- x@horizons
```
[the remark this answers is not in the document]
brownag: It is actually done down below

R/SoilProfileCollection-operators.R
brownag: I think that any time you do anything that could possibly be misinterpreted by data.table you need to call the method that will ensure that the data.table code is triggered on a data.table object. 

It is possible, though unlikely, that the metadata could say data.table but the user or another operation converted to data.table. Wrapping access to horizons in a call when you get it out and when you put it back in is the best way to ensure that.

brownag: This back and forth I did here is weird. I think I was having an unrelated error and tried fiddling with this and did not change it back. Thanks for drawing my attention to it.  I can clean it up and call it when the slots (site and horizon) are accessed. This is all to support data.table. tbl_df is essentially a drop in replacement for data.frame.

R/Class-SoilProfileCollection.R
```diff
@@ -51,19 +51,7 @@ setClass(
...
-    df_class_correct <- all(unlist(lapply(df.slots, function(slt) {
-      # only enforce df class if the metadata are set
-      daf <- slot(object, slt)
-      if (length(mdf) == 1 & class(daf)[1] != mdf)
-        slot(object, slt) <- .as.data.frame.aqp(daf, mdf)
-      return(TRUE)
-    })))
-
-    return(spc_in_sync(object)$valid & df_class_correct)
+    return(spc_in_sync(object)$valid)
```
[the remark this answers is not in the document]
brownag: I was testing whether I could have it there. I do things in branches and undo them -- I try lots of things and break lots of things. But I think the main issue I removed it is that it doesnt work. setting slots that way, using slot<- was not persisting to the object -- which of course makes sense. I wasn't thinking -- and it basically never met the conditions to trigger in the first place.

R/SoilProfileCollection-operators.R
dylanbeaudette: I'll have to think about this some more, during daylight hours. That is an incredibly verbose chunk of code required to perform some very basic operation on a `data.frame`-like object.  I'm starting to wonder if `data.table` would be more useful for on-the-fly operations vs. drop-in replacement. 

An example, would the internals of e.g. `horizonColorIndices` have to be re-written using `as.data.frame.aqp` and / or `.data.frame.j`?

R/Class-SoilProfileCollection.R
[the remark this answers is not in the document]
dylanbeaudette: OK. For future review: should I interpret a PR as "pleas take a look at these things", or "please wait until I ask you to look at these things"?

R/SoilProfileCollection-operators.R
brownag: In that example, `horizons<-` handles it when merge `horizonColorIndices` back into the SPC

`horizonColorIndices` is written in a way that is unambiguous. `.data.frame.j` is only required for multi column subsetting in the j position. .as.data.frame.aqp is not needed either, because it is not a method that is internal to the object. Only the objects methods care about maintaining the data.frame object state. Random other methods do not care. They may take advantage of whatever class is there, or even convert to one, but I think that there is not as much "overhead" here as you are thinking. The major barrier is my obtuse code, which I will ponder how I can make clearer and document for you.

brownag: Crap I deleted my comment. Here is the issue. 

horizonColorIndices will actually break if the object in the SPC is a data.table. It is unambiguous up till the last line.

```
  return(hz[,c(idname(p), hzidname(p), names(hz)[!names(hz) %in% names(horizons(p))])])
```
needs to be
```
  return(.data.frame.j(hz. c(idname(p), hzidname(p), names(hz)[!names(hz) %in% names(horizons(p))])))
```

brownag: However, any potential conversion would happen internal to the object in e.g. `horizons<-` when the color indices were reinserted. The horizon slot should already be the same class, so only the class is checked internally by `.as.data.frame.aqp`

If you think we can switch over to data.table entirely I am willing to see what of these non-central methods will break -- maybe try changing the SPC object definition in a branch? I do not run the full panel of aqp tests on data.tables, so surely there are many more things that will break in the less exercised parts of the codebase

R/Class-SoilProfileCollection.R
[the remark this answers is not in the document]
brownag: I will use the draft PR request feature, and convert to full PR when ready, in the future. Once I move it over, I want you to look at it. I had other things going on today and did not get around to pushing the last commits that I had needed. Clearly the min/max changes needed some work ... but the earlier commits were fine-ish.

Im ready for you to review -- thanks for catching a couple things I missed in those most recent experiments with min and max. I personally think tapply is fine there and didnt really even think a data.table implementation was necessary until i tried the show method on a million profiles. Takes like 12-14 seconds.

R/SoilProfileCollection-operators.R
brownag: The above is now fixed for horizonColorIndices.

brownag:
I think all of the pending review items have been resolved thus far. Is there anything else you want done on this PR?

Issue #34, opened 2017-07-20
re-make munsell LUT with 2.5 values via interpolation


dylanbeaudette:
Need to post / archive / update the old code used to generate `munsell.rda`. That file was generated using a combination of raw data and manually added neutral (N) hues.

 * `munsell` package has a nice script for using the raw data and integrating greys
 * greys can be found here: http://wiki.laptop.org/go/Munsell

dylanbeaudette:
related to #17, Munsell LUT files added to `/misc/utils/Munsell`

dylanbeaudette:
This will be properly fixed by interpolation of all 1/2 chips (#177)

PR #177, opened 2021-01-05
All pairwise dE00 contrasts for integer `aqp::munsell` "chips" -> "equivalent" Munsell chips lookup list

brownag:
I wanted to tabulate chips that were "perceptually equivalent" -- I realized that these high chroma chips were very very similar    and I wondered how this type of thing affects "uniqueness" of rgb to munsell chip conversions. This is looking for chips that are so dang close to being perceptually the same but have subtle numerical deviations.

EDIT: add example of different very red chips that are essentially the same (veeeeery subtle differences). 
image

ALSO: these birds have such inspiring colors whose uniqueness I want to quantify
image

Current estimates of the [arbitrary] threshold are probably too low 
 - [x] mask out diagonals -- does this compare with literature for dE00 values?
 - [x] "distance" matrix returned by compare_colour() is not the same as typical `dist` objects -- investigate @dylanbeaudette 
 - [x] get a couple of reasonable dE00 values that map to the limits of human perception for thresholds and pick one or a range: e.g. dE00 4, 6, 10.


dylanbeaudette:
We can safely merge this before all TODO items are determined. `getClosestMunsellChip` has a similar name but a more specific manual page would remove ambiguity.

dylanbeaudette:
Discussion of dE00 thresholds that I have found useful:
 * http://zschuessler.github.io/DeltaE/learn/
 * http://www2.ece.rochester.edu/~gsharma/ciede2000/ciede2000noteCRNA.pdf
 * https://onlinelibrary.wiley.com/doi/abs/10.1111/1556-4029.14193 (I can send a copy if you'd like)

dylanbeaudette:
Here is how I typically generate a distance matrix using `farver::compare_colour()`:

```r
# dE00
# pair-wise distances are only present in the upper triangle
# lower triangle is 0
d <- compare_colour(x[, c('L', 'A', 'B')], from_space='lab', white_from = 'D65', method='cie2000')
# copy color codes for convenience
dimnames(d) <- list(x$munsell, x$munsell)

# convert to dist object, which expects distances in the lower triangle
d.dist <- as.dist(t(d))
```
There are of course other ways to do this.

brownag:
Thanks for all the great references and the pointers on the distance matrix. I read over them and couldn't resist continuing on this a bit. But I will basically be handing it over to you now to play with and we can merge in as you like.

I am always needing reminding of the non-uniform changes from "chip to chip" depending on the direction of that change in Munsell space. I agree we need a way to integrate info on chip adjacency into the dE00 matrix aggregation. Currently all "different" chips are assigned _equal weight_ in my quantile calculation, BUT we know that some changes in "whole chip" are more different than others -- certain pair-wise contrasts are expected to be more "perceptible"

I like [this one](https://zschuessler.github.io/DeltaE/learn/) that says a difference of "1 to 2" dE00 is "perceptible by close observation"

Pretty consistent with what I see with a threshold of **2.15** (global 0.001 quantile from aqp::munsell dataset). If I squint my eyes or look at the screen at an angle, it can tell at least some sRGB colors in a "equivalent" set on screen are different. Er.. maybe 

image

For lower limit perceptibility, I think something around dE00 ~2 fits the concept for what this algorithm is (or I intended). The FBI use a low end of range ~3.5, per the paper you posted for identifying similar soil materials -- so I can't imagine our concept of "practically/perceptually" equivalent going much higher than 2 or 3 -- even if we do come up for a way to better determine that, and even if the average change in chip is something higher than that. 

The other higher levels you suggest would also be interesting to have handy. Perhaps this list I have made could store list of numeric matrix rather than vector? Could definitely use that for grouping sets of colors at varying granularity. I don't know that it can completely replace your more generic cases involving calling {farver} on the fly. 

Taking the upper triangle definitely improves things for my naive approach. Your suggested/typical syntax is cleaner and it is 2x faster (30s versus 1 minute). However makes it slightly more complicated to do the indexing for lookup in `munsell` and I didnt quite figure it out. TODO in there, have at trying to fix it/make it faster.

Masking out diagonals and the lower triangle gives the following results for the "global" quantiles (at low probability levels of interest to identifying whole chips that are "perceptually equivalent" to some other given whole chip). 

```
> quantile(x, p = c(0.001,0.002,0.003,0.004,0.005,0.01), na.rm = TRUE)
    0.1%     0.2%     0.3%     0.4%     0.5%       1% 
2.145926 3.359188 4.185346 4.865601 5.447339 7.636058 
```

And a visual way of comparing how arbitrary chosen probability of 0.001 works:
image

Also for more visual ideas, I updated the OP with some inspirational pictures.

dylanbeaudette:
> RE: getClosestMunsellChip

After thinking about this some more, there really isn't an issue with ambiguity here. If anything, I'd prefer to re-name this function to `closestMunsellChip`. The entire point of that function is to consistently "snap" a non-standard Munsell notation to a reasonable approximation in our `munsell` database, without invoking another dependency or attempting to re-implement the complex math in `munsellinterpol`.

dylanbeaudette:
> For lower limit perceptibility, I think something around dE00 ~2 fits the concept for what this algorithm is (or I intended). The FBI use a low end of range ~3.5, per the paper you posted for identifying similar soil materials -- so I can't imagine our concept of "practically/perceptually" equivalent going much higher than 2 or 3 -- even if we do come up for a way to better determine that, and even if the average change in chip is something higher than that.

I'm glad you found these references useful. I had dE00 of *~4* in mind, but something between 2-3 is closer to your original objective. Fine by me as long as we remind ourselves that:
  * we are working with field data most of the time
  * our monitors are not calibrated (precision / calibrated monitors are dE00<= 2)
  * out lighting is not controlled
  * older literature (SSSAJ Special Pub. 21) suggests that skilled soil scientists under consistent conditions are precise to about "1/2 chip" in our color books--that translates to a dE00 of 2-4.

As a side note, all this tinkering / thinking caused me to notice a bug in `contrastChart` when using very low dE00 thresholds.

brownag:
Cool, I like all that. And I am totally in agreement about those reminders -- in fact, I think identifying the practical lower limits of color perception and precision (in terms of chips that we think about) are essential in light of those things, and that's a good way of wording some of my underlying motivation for doing this.

Do you agree with my use of `equivalent` in this case as being "practically" equal but not exactly numerically equal (in terms of actual color coordinates in any of our systems.  If you swapped a 7.5YR 3/2 for a 5YR 3/2 in an image, or on a ped face -- would you really notice a meaningful difference? You may be able to tell the difference between a chip and its equivalent chips when they are side by side but they would "function" the same. 

I originally was thinking `similar` but then I felt that `similar` is too broad for my low-probability stuff -- broader sort of necessitates some options for customizing the threshold.

Also, regarding interpolation -- have you considered augmenting the lookup table with just the handful of half-chips we have in the standard soil Munsell book? Those low-value/low-chromas on some pages e.g. 7.5YR. 

I like the integers approach for the full LUT, but would a subset of `munsell`, or a method that subsets/modifies it for just chips in the book be useful?

dylanbeaudette:
> Taking the upper triangle definitely improves things for my naive approach. Your suggested/typical syntax is cleaner and it is 2x faster (30s versus 1 minute). However makes it slightly more complicated to do the indexing for lookup in munsell and I didnt quite figure it out. TODO in there, have at trying to fix it/make it faster.

I'll look through my notes on how I typically use the results: transpose→lower triangle, distance object, coercion to matrix object.

brownag:
> > Taking the upper triangle definitely improves things for my naive approach. Your suggested/typical syntax is cleaner and it is 2x faster (30s versus 1 minute). However makes it slightly more complicated to do the indexing for lookup in munsell and I didnt quite figure it out. TODO in there, have at trying to fix it/make it faster.
> 
> I'll look through my notes on how I typically use the results: transpose→lower triangle, distance object, coercion to matrix object.

I tried that, and a couple other things. But was getting garbage chips. The order was getting messed up, but stats were right. I'd welcome any suggestions, cause in theory you should only need that upper triangle then you can recycle the distances to lower to do the indexing I do. And perhaps there is a better way you can derive the index positions of all the chips -- but like I said, it was too much for me to figure out (late at night)

dylanbeaudette:
> Cool, I like all that. And I am totally in agreement about those reminders -- in fact, I think identifying the practical lower limits of color perception and precision (in terms of chips that we think about) are essential in light of those things, and that's a good way of wording some of my underlying motivation for doing this.
>
> Do you agree with my use of `equivalent` in this case as being "practically" equal but not exactly numerically equal (in terms of actual color coordinates in any of our systems. If you swapped a 7.5YR 3/2 for a 5YR 3/2 in an image, or on a ped face -- would you really notice a meaningful difference? You may be able to tell the difference between a chip and its equivalent chips when they are side by side but they would "function" the same.

Yes, `equivalent` is a good choice. Elaboration / demonstration of this phenomenon would help explain artifacts of other color functions in {aqp}, `aggregateSoilColor`, `colorQuantiles`, `mixMunsell`, etc. where the resulting colors may include hues not present in the source data. It would also help to limit the huge list of chips specified in a RIC.
 
> I originally was thinking `similar` but then I felt that `similar` is too broad for my low-probability stuff -- broader sort of necessitates some options for customizing the threshold.

Agreed. Nit-pick: I suggest low-contrast vs. low-probability.

> Also, regarding interpolation -- have you considered augmenting the lookup table with just the handful of half-chips we have in the standard soil Munsell book? Those low-value/low-chromas on some pages e.g. 7.5YR.
> I like the integers approach for the full LUT, but would a subset of `munsell`, or a method that subsets/modifies it for just chips in the book be useful?

Good idea, yes. I have considered it (the interpolation isn't all that difficult) but so far there hasn't been a good use case. This might be it. I added the 2.5 chips in #34, but made a new issue to explore this further (#178). Perhaps an additional object `munsell.half` that covers the soil color book pages to the "half-chip". If this is useful, then we could interpolate the Munsell reference spectra to the 1/2 chip.

brownag:
> low-contrast vs. low-probability.

Fair point; I caught myself a few times RE not saying "contrast", I genuinely prefer referring to the {farver} dE00 pair-wise "differences" as contrasts and will continue to try to remember to do this thanks to your careful wording.

Counter-nitpick:

Probability in the general sense is not "wrong" here: `equivalentMunsellChips` uses a lookup table based on dE00 contrast thresholds user-defined by low probability (p=0.001) within the "population" of **all** contrasts between whole-number "chips" in the `aqp::munsell` dataset.

brownag:
> Good idea, yes. I have considered it (the interpolation isn't all that difficult) but so far there hasn't been a good use case. This might be it. I added the 2.5 chips in #34, but ...

```
> data("munsell")
> levels(factor(munsell$value))
 [1] "1"  "2"  "3"  "4"  "5"  "6"  "7"  "8"  "9"  "10"
```

This is a really good idea. I was initially confused, because the 2.5 chips arent in `munsell` the dataset, but https://github.com/ncss-tech/aqp/tree/master/misc/utils/Munsell

> made a new issue to explore this further (#178). Perhaps an additional object `munsell.half` that covers the soil color book pages to the "half-chip". If this is useful, then we could interpolate the Munsell reference spectra to the 1/2 chip.

This is a great idea!

dylanbeaudette:
> > low-contrast vs. low-probability.
> 
> Fair point; I caught myself a few times RE not saying "contrast", I genuinely prefer referring to the {farver} dE00 pair-wise "differences" as contrasts and will continue to try to remember to do this thanks to your careful wording.
> 
> Counter-nitpick:
> 
> Probability in the general sense is not "wrong" here: `equivalentMunsellChips` uses a lookup table based on dE00 contrast thresholds user-defined by low probability (p=0.001) within the "population" of **all** contrasts between whole-number "chips" in the `aqp::munsell` dataset.

True. Which do you think conveys the simplest practical interpretation? Rarely in what we do is there right/wrong, but shades of usefulness. I find that incorporating "contrast" into our discussion of color is a simple way to connect with color science literature and more clearly define an objective.

dylanbeaudette:
> > Good idea, yes. I have considered it (the interpolation isn't all that difficult) but so far there hasn't been a good use case. This might be it. I added the 2.5 chips in #34, but ...
> 
> ```
> > data("munsell")
> > levels(factor(munsell$value))
>  [1] "1"  "2"  "3"  "4"  "5"  "6"  "7"  "8"  "9"  "10"
> ```
> 
> This is a really good idea. I was initially confused, because the 2.5 chips arent in `munsell` the dataset, but https://github.com/ncss-tech/aqp/tree/master/misc/utils/Munsell

Crap, I made the issue then incorporated an ugly work-around. Thanks for reminding me!
```r
## 2016-03-07: "fix" values of 2.5 by rounding to 2
  the_value <- ifelse(the_value == 2.5, 2, the_value)
```
#178 is the correct "fix".

> > made a new issue to explore this further (#178). Perhaps an additional object `munsell.half` that covers the soil color book pages to the "half-chip". If this is useful, then we could interpolate the Munsell reference spectra to the 1/2 chip.
> 
> This is a great idea!

I'll get to this when I can, still a huge pile.

brownag:
> > > low-contrast vs. low-probability.
> > 
> > 
> > Fair point; I caught myself a few times RE not saying "contrast", I genuinely prefer referring to the {farver} dE00 pair-wise "differences" as contrasts and will continue to try to remember to do this thanks to your careful wording.
> > Counter-nitpick:
> > Probability in the general sense is not "wrong" here: `equivalentMunsellChips` uses a lookup table based on dE00 contrast thresholds user-defined by low probability (p=0.001) within the "population" of **all** contrasts between whole-number "chips" in the `aqp::munsell` dataset.
> 
> True. Which do you think conveys the simplest practical interpretation? Rarely in what we do is there right/wrong, but shades of usefulness. I find that incorporating "contrast" into our discussion of color is a simple way to connect with color science literature and more clearly define an objective.

I don't think there is a dichotomy, but I agree on finding more ways to incorporate "contrast" in discussion. 

To me, the probability threshold is easy to explain -- you take all the chips and contrast them with all other chips, that is your sample -- you define some very small, but consistent threshold such as 0.001. This type of thing is commonly done for e.g. hydrologic design. The actual threshold value of dE00 _contrast_ depends on the specific input data, quantile method etc. -- there are many possible ways to specify thos -- but the concept of a small probability corresponds with a comparably small contrast essentially no matter what your method is. That is just my thought, and I was simply referring to how my method worked by using low probability. I suppose we could just hard-code a 2 in there and cite some other reference.

brownag:
This branch is safe to rebase and merge -- that would be my preferred strategy for something like this that is independent of existing functionality (aside from depending on `munsell` dataset) -- per our discussions today.

I request a final review of it @dylanbeaudette, please just scan over it. You don't have to commit changes to the https://github.com/ncss-tech/aqp/tree/similarChip branch, but I am done with it now -- so feel free to wordsmith documentation with any verbiage you thing would be a clearer way to convey the relevant concepts!

brownag (commented):
Minor tweaks to docs, still not perfect, but tried to be "complete"

dylanbeaudette (commented):
I think that we should merge this. Thanks for all of the hard work and great idea. I'll do some final word-smithing (maybe) later this week.

brownag:
OK, thanks. I am still somewhat wondering whether that 2.15 is "too conservative" -- but I would rather lean that way than the other in the realm of calling chips "equivalent"

PR #183, opened 2021-01-17
mixMunsell patch

brownag:
I noticed a couple small things -- edge cases -- while messing around with the Harden color index functions on the `jacobs2000` data -- added some tests to the master branch and I fix them here -- this is the one conflicting file.

### DIFF
--- a/R/mixMunsell.R
+++ b/R/mixMunsell.R
@@ -119,7 +119,7 @@ mixMunsell <- function(x, w = rep(1, times = length(x)) / length(x), n = 1) {
 
   # can't mix a single color, just give it back at 0 distance
   if (length(unique(x)) == 1) {
-    return(data.frame(munsell = x, distance = 0))
+    return(data.frame(munsell = x[1], distance = 0))
   }
 
   # must have as many weights as length of x
--- a/R/mixMunsell.R
+++ b/R/mixMunsell.R
@@ -3,36 +3,36 @@
 
 # helper function for printing out value / chroma ranges by hue
 .summarizeMunsellSpectraRanges <- function() {
-  
+
   # make R CMD CHECK happy
   munsell.spectra <- NULL
-  
+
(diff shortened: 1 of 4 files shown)


R/mixMunsell.R
```diff
@@ -78,66 +78,97 @@
...
-    
-  # x and w same lengths
-  
-  # x contains valid Munsell
-  if(any(is.na(parseMunsell(x)))) {
-    stop('input must be valid Munsell notation, neutral hues not supported')
+
+  # can't mix a single color, just give it back at 0 distance
+  if (length(unique(x)) == 1) {
+    return(data.frame(munsell = x, distance = 0))
```
brownag, line 122: `unique(x)`

R/mixMunsell.R
```diff
@@ -149,52 +180,54 @@ mixMunsell <- function(x, w = rep(1, times = length(x)) / length(x), n = 1) {
...
     munsell = names(m.match),
     distance = m.match,
     stringsAsFactors = FALSE
   )
-  
+
   # clean-up row names
-  row.names(res) <- as.character(1:nrow(res))
-  
+  row.names(res) <- NULL
```
dylanbeaudette, line 228: Hey that is cool, learned something new.

dylanbeaudette (approved):
Thanks for improving this function and tests.

R/mixMunsell.R
brownag: It is cool, save a couple key strokes. 

Not that it matters for this function but non-`NULL` `row.names<-` to a [`tibble` causes warnings]( https://tibble.tidyverse.org/reference/rownames.html). I think the thought is that they get do get messy, as above, and if they contain any real info they should be a column.

dylanbeaudette: Good point. I'll be sure to use this construct moving forward.

PR #186, opened 2021-01-19
allowing for MD syntax in roxygen blocks

### LINKED ISSUE #184: Allow for MD style authoring of Rd content
MD-style authoring of manual pages is super convenient, but will break a lot of documentation.

https://cran.r-project.org/web/packages/roxygen2/vignettes/rd-formatting.html

Enable via toggling:
```
# in DESCRIPTION
Roxygen: list(markdown = FALSE)
```

dylanbeaudette:
After a review, this should close #184


brownag:
I did a Find in Files with regex: `^#'.*\[.*\]` and looked at a bunch of Roxygen v.s. Rd v.s. rendered. It appears that everything problematic about the semantics of `[` for markdown are converted over to use ` `` `, in a code block (examples or usage) or at least properly escaped e.g. `\[0, 1]`

I am particularly excited about doing things like `[glom()]` or `[fetchNASIS in soilDB][soilDB::fetchNASIS()]` for linking instead of `\code{\link{glom}}` etc. 

There are lots of places where `\code`/`\link` are still used -- a few hundred lines, many with several, possibly nested uses per line -- fun for another PR? Or worth tackling here? 

Otherwise I think this is good to merge.

brownag (approved):
Passes `devtools::check_man()`
remaining instances of [ in roxygen comments appear to be properly handled
spot checked and appear to render correctly. 

May need to check the rendered HTML to verify everything looks good there, but the .Rd files generated seem fine.

Issue #174, opened 2020-12-04
data.table rewrite of logic functions, profileApply / mutate_profile


brownag:
This is a sub-issue related to #157 -- which is too "big" to tackle in one bite IMO.

Significant benefits can come from `data.table` rewrite of logic functions and soon `mutate_profile` / `profileApply` in terms of memory usage and elapsed time for profile-level evaluations. For this `sp4`-based demo, the data.table rewrite is ~50x faster in fast mode (most commonly used output) and 10x faster in slow mode (identical output).

Here is demo showing a drop-in-replacement for checkHzDepthLogic.

``` r
library(data.table)
library(aqp, warn.conflicts = FALSE)
#> This is aqp 1.26

data(sp4)
depths(sp4) <- id ~ top + bottom

aqp_df_class(sp4) <- "data.table"
sp4 <- rebuildSPC(sp4)
#> using `hzID` as a unique horizon ID
```

New function

``` r
#' Fast checks of horizon depth logic in entire SoilProfileCollection
#' @param object A SoilProfileCollection
#' @param fast If details about specific test results are not needed, the operation can run approximately 5x faster. Default: TRUE.
#' @return A data.table containing profile IDs, validity boolean and test results if 
#' @export

checkAllLogic <- function(object, fast = TRUE) {
  h <- horizons(object)
  hzd <- horizonDepths(object)
  
  hby <- substitute(idname(object))
  top <- substitute(hzd[1])
  bottom <- substitute(hzd[2])
  
  if (!fast) {
    res <- h[, .(tests = list(tests = data.frame(t(hzDepthTests(eval(top), eval(bottom)))))), by = c(eval(hby))][ 
             , .(tests = tests, valid = all(!tests[[1]])), by = c(eval(hby))]
    res <- cbind(res, rbindlist(res$tests))
    res$tests <- NULL
    return(res)
  } else {
    res <- h[, all(!hzDepthTests(eval(top), eval(bottom))), by = c(eval(hby))]
    colnames(res) <- c(idname(object), "valid")
    return(res)
  }
}
```

"Fast" results

```r
checkAllLogic(sp4)
#>                 id valid
#>  1:         colusa  TRUE
#>  2:          glenn  TRUE
#>  3:          kings  TRUE
#>  4:       mariposa  TRUE
#>  5:      mendocino  TRUE
#>  6:           napa  TRUE
#>  7:     san benito  TRUE
#>  8:         shasta  TRUE
[...]

Issue #157, opened 2020-07-15
Phase-in of data.table support in SPC objects and functions returning SPC objects


dylanbeaudette:
Keeping track of changes / testing related to `data.table` replacement of `plyr` / `reshape` functionality. Each change should be checked and a new test added if there isn't sufficient coverage. See related issue in [soilDB](https://github.com/ncss-tech/soilDB/issues/161) and [sharpshootR](https://github.com/ncss-tech/sharpshootR/issues/32).

### `join` → `base::merge(..., `incomparables = NA`, sort = FALSE)` or `data.table` join
 - [x] `SoilProfileCollection-coercion`
 - [x] `SoilProfileCollection-slice-methods` (deprecated, in favor of `dice`)
 - [x] `munsell2rgb`

### `ddply`
 - [x] `SoilProfileCollection-setters`
 - [x] `aggregateColor`
 - [x] `aggregateSoilDepth`
 - [x] `getmlhz`
 - [x] `missingDataGrid`

### `llply`
 - [x] `profile_compare`

### `ldply`
 - [x] `SoilProfileCollection-slice-methods` (deprecated, in favor of `dice`)
 - [x] `aggregateColor`
 - [x] `soilColorSignature`

### `dlply`
 - [x] `aggregateColor`
 - [x] `profile_compare` (deprecated, in favor of `NCSP`)

### `cast` → `data.table::dcast`
 - [x] `evalGenHz`
 - [x] `soilColorSignature`

Note that `value` argument becomes `value.var`.

### `melt` → `data.table::melt`
 - [x] `plotColorQuantiles`
 - [x] `evalGenHz`
 - [x] `slab`
 - [x] `slab` examples will need `reshape` package alternatives
 - [x] `soilColorSignature`

If the object is not a `data.table` then perform on-the-fly conversion to/from. The following works in `plotColorQuantiles`:

```r
as.data.frame(
data.table::melt(
as.data.table(res$marginal), 
id.var = c('p', 'L_colors', 'A_colors', 'B_colors', 'L_chip', 'A_chip', 'B_chip')
)
)
```
However, there following appears with a similar conversion in `evalGenHz`:
```
In melt.data.table(as.data.table(h), id.vars = genhz, measure.vars = c(vars,  :
[...]

PR #190, opened 2021-01-20
Implement data.table checkHzDepthLogic RE #157, #174

brownag:
Addresses my first demonstration point from #174 

RE: #174 I am thinking that full generalization of `profileApply` is not worth it -- I have some ideas on `mutate_profile`. That should perhaps be spliced off into a new, back-burner issue.


dylanbeaudette (approved):
Looks good to me. I have not tested, but I am sure that you have. Learned about `eval()`. I'll add this to my recent dive into `data.table`. Any word on how much faster this is?

brownag:
See #174 for the benchmarks -- it is ~5-10x faster

brownag:
Sorry, mis-read my own bench mark. Raw data.table is >50x faster in my benchmark. But there is some overhead from the way the function is set up.

PR #192, opened 2021-01-21
site normalization fixes

brownag:
Closes #171 (which still needs a working example, see #172), also uses `data.table` .SD style syntax to replace iteration by `ddply` #157.

Adds a new warning if you specify a problematic formula to `site<-` -- i.e. one that does not produce a result that is 1:1 with the existing site table / profile ID / `profile_id()`-- leaving data unchanged. 

This is hopefully OK and sufficient to explain to users who might have expected the horizon var to move to site.

Also, I believe it adds 7 tests which should put us... over 1000!

Issue #171, opened 2020-11-18
site(d) <- results in corrupt profiles


dylanbeaudette:
For some reason the NEON data once converted into an SPC does not play nicely with:
```r
subset(S, siteID %in% c('SJER', 'SOAP', 'TEAK'))
```
There is a lot of "magic" involved with getting the NEON data into R, resulting in a global environment full of objects. Also, the final object `S` contains many duplicated columns resulting from joins.

This has nothing to do with `aqp::subset`, rather novel input data and `site(x) <-`.

Setup:
```r
# Load required packages
library(neonUtilities)
library(aqp)
library(dplyr)

## ----download-data, results='hide'--------------------------------------------------------------------

MP <- loadByProduct(dpID = "DP1.00096.001", check.size = F)

# Unlist to environment - see download/explore tutorial for description
list2env(MP, .GlobalEnv)



## ----join-bgc-----------------------------------------------------------------------------------------

# duplicate the 'horizon' information into a new table
S <- mgp_perhorizon

# duplicate the biogeochemical information into a new table
B <- mgp_perbiogeosample

# Select only 'Regular' samples (not audit)
B <- B[B$biogeoSampleType=="Regular" & 
         !is.na(B$biogeoSampleType),]

# Join biogeochem data to horizon data
S <- left_join(S, B, by=c('horizonID', 'siteID',
                          'pitID','setDate',
                          'collectDate',
                          'domainID',
                          'horizonName'))
S <- arrange(S, siteID, horizonTopDepth)



## ----site-labels--------------------------------------------------------------------------------------

## combine 'domainID' and 'siteID' into a new label variable
S$siteLabel=paste(S$domainID, S$siteID, sep="-")


# init SPC
depths(S) <- siteLabel ~ horizonTopDepth + horizonBottomDepth
```

Corrupt SPC here:
```r
# move site-level attributes to @site
site(S) <- ~ siteID + pitNamedLocation.x + pitID + labProjID + laboratoryName + nrcsDescriptionID
```

View error here:
```
S[42, ]
```

brownag:
Narrowed the error condition (SPC validity method returns FALSE) down to inclusion of `labProjID` and/or `laboratoryName` in the `site<-` (formula) normalization call

That is: 

```
site(S) <- ~ siteID + pitNamedLocation.x + pitID + nrcsDescriptionID
S[42,]
```
works as expected.

[...]

PR #194, opened 2021-01-22
cleanup .createSiteFromHorizon

brownag:
#171 #192 -- is iteration needed? -- No, but this PR cleans up comments etc and demonstrates .SD / .SDcols / by / j syntax and adds one slightly less trivial test

### DIFF
--- a/R/SoilProfileCollection-setters.R
+++ b/R/SoilProfileCollection-setters.R
@@ -339,14 +339,12 @@ setReplaceMethod("site", signature(object = "SoilProfileCollection"),
   # the horizon data !
   idx <- idx[-match(idname(object), names_attr)]
 
-  # this will break when multiple horizons in the same pedon have different site data!
-  # this seems to work fine in all cases, as we keep the ID column
-  # and it ensures that the result is in the same order as the IDs
   .SD <- NULL
   
   dth <- as.data.table(horizons(object))
+  idn <- dth[[idname(object)]]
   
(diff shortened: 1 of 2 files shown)


R/SoilProfileCollection-setters.R
```diff
@@ -339,14 +339,12 @@ setReplaceMethod("site", signature(object = "SoilProfileCollection"),
...
-  # this will break when multiple horizons in the same pedon have different site data!
-  # this seems to work fine in all cases, as we keep the ID column
-  # and it ensures that the result is in the same order as the IDs
   .SD <- NULL
   
   dth <- as.data.table(horizons(object))
+  idn <- dth[[idname(object)]]
   
-  new_site_data <- .as.data.frame.aqp(unique(dth[, .SD, .SDcols = names_attr]), aqp_df_class(object))
+  new_site_data <- .as.data.frame.aqp(dth[, unique(.SD), .SDcols = names_attr, by = list(idn)], aqp_df_class(object))
```
brownag, line 347: Now third-or-fourth guessing this -- I think the current code is safe -- since the profile ID (`idname(object)`) is part of `names_attr`... but this is how you would apply unique to each profile in the horizon data to make site.

brownag: So, I guess this is just syntax comparison -- apply `unique()` to whole data.table (where idname is a column) versus apply unique on profile-horizon subdataframes with `j =unique(.SD)`

PR #195, opened 2021-01-22
Rd2roxygen #139 + more

brownag:
This was an experiment. Should pass checks but needs more human review. At one point I double `cat`'d the roxygen out to template files, so there were some duplicates of the same roxygen block. I think I caught them all ...

**Edit: the true "test" is delete the whole aqp /man folder, then have roxygen2 rebuild it. Should work. Except for NAMESPACE.**

Rd2roxygen output by file

new .R files

resolve existing .R files

properly converted .Rd files

fix for check_man

cleanup


R/argillic.clay.increase.depth.R
```diff
@@ -0,0 +1,29 @@
+#' Return upper boundary of argillic horizon
```
brownag, line 1: whoops, missed moving this one to getArgillicBounds.R

brownag (commented):
went over them all

PR #203, opened 2021-02-19
Remove {rlang} dependency / rename/deprecate some dplyr-like methods

brownag:
This is a demonstration of making all {aqp} non-standard evaluation currently in use rely on {base} rather than {rlang}

It also addresses several problematic function names that result in namespace conflicts with {stats}, {dplyr}, etc.


brownag:
And regarding an answer to my question in my commit message https://github.com/ncss-tech/aqp/pull/203/commits/32476c843ba1afd74f86656e4fa595fbc9c2e400: I believe the hitch with `subset,SoilProfileCollection-method` and needing `parent.frame(n=2)` has to do with how it is being dispatched via S4 and/or going through the `base::subset` method, making it "one level deeper" in the call stack relative to the calling environment than the other methods are.

dylanbeaudette:
Works for me. I like the idea of fewer namespace conflicts and a leaner package. The only NSE that I currently use / maintain is managed with base.

brownag:
I think this PR is safe to merge, but I am thinking that deprecating all of these function names should warrant a version bump. I know we just recently bumped to 1.28, but that was overdue since CRAN submission.

dylanbeaudette:
Just tested with a couple of tutorials, seems fine.

brownag:
OK. Here is the case I think I was concerned about, also seems to work fine.
``` r
library(aqp, warn=FALSE)
#> This is aqp 1.29
library(soilDB)
data("loafercreek")

anUncertainCase <- function(x) {
  .moreUncertain <- function(y) {
    subset(y, taxonname != "Loafercreek")
  }
  .moreUncertain(x)
}

groupedProfilePlot(anUncertainCase(loafercreek), groups = "taxonname")
```

PR #206, opened 2021-02-21
Moving various Soil Taxonomy heuristics to sharpshootR

brownag:
This moves various Soil Taxonomy heuristic functions to sharpshootR (see https://github.com/ncss-tech/sharpshootR/pull/36 for the other side of this PR)

Lays out a basic scheme (see R/moved-to-sharpshootR.R) for _temporary_ redirection of function names to sharpshootR--which is in {aqp} suggests--and generally is present in users library. Therefore, the functions can generally be found and dispatched even if called from {aqp} without sharpshootR pre-loaded. I tested this in a variety of different configurations and it seems to work / be valid for sharpshootR in suggests. I considered emitting a message, but that is annoying/extra book keeping to keep it from being annoying.

Standard ways of loading the packages eg. `library(aqp)` followed by `library(sharpshootR)` should avoid any extra overhead from the methods being dispatched thru {aqp}.

The temporary redirection might be an unnecessary step -- I could just remove the functions -- but this seemed like a decent way to make folks aware and maintain some back-compatibility.


brownag:
This was a fine proof of concept for Suggests-dependencies--such as functions moved around within the AQP "metapackage"--including soilDB, sharpshootR. After discussion with @dylanbeaudette and thinking since created, there is little value at this point to changes in where these functions are. 

I would like to harmonize these these types of applied approaches (currently focused around `profileApply`-like workflows) and provide some sort of a generic interface that can both leverage intermediate results efficiently, work on large collections, and provide a simple high-level user API. Looking for ideas on how to best do this!

Issue #205, opened 2021-02-19
`findHzGaps()`, `fillHzGaps()` function to find / fill gaps in horizon depths


dylanbeaudette:
Need a function to reliably "fill" horizon gaps in existing data or as a product of filtering on `checkHzDepthLogic(..., byhz = TRUE)

 - [x] `fillHzGaps()`
 - [ ] ~`findHzGaps()`~
 - [x] `data.table` optimization

Example using latest {aqp}.
```r
library(aqp)
library(data.table)

d <- lapply(letters[1:10], random_profile, n = c(6, 7, 8), n_prop = 5, method = 'LPP', SPC = TRUE)
d <- combine(d)

# sprinkle NA
z <- d
z$bottom[2] <- NA
z$top[20] <- z$bottom[20]

z$bottom[32] <- 15
z$bottom[5]
z$top[6] <- 95

# dropping horizons (a good idea?)
zz <- dice(z, byhz = TRUE)

# ok
metadata(zz)$removed.horizons

par(mar = c(0, 0, 0, 1))
plotSPC(zz, name = NA, divide.hz = FALSE, default.color = 'grey')
```

![image](https://user-images.githubusercontent.com/624277/108555817-d9dbce80-72aa-11eb-8206-20f09b9938e2.png)

dylanbeaudette:
Functional, tested, documented (but not optimized) `fillHzGaps` now in master branch.

brownag:
Here is an attempt--some 3x faster on `sp4`--using only {base}.

``` r
library(aqp, warn = FALSE)
#> This is aqp 1.27

# original code, slightly modified relative to master
fillHzGaps <- function(x, flag = FALSE) {
  
  idn <- idname(x)
  hzidn <- hzidname(x)
  htb <- horizonDepths(x)
  hznames <- horizonNames(x)
  
  ids.top.bottom.idx <- match(c(idn, hzidn, htb), hznames)
  
  h <- horizons(x)
  
  hs <- split(h, h[[idn]])
  
  h.filled <- lapply(hs, function(i) {
    n <- nrow(i)
    
    s <- 1:(n-1)
    
    .top <- i[[htb[1]]]
    .bottom <- i[[htb[2]]]
    idx <- which(.bottom[s] != .top[s + 1])
    
    if(length(idx) > 0) {
      gap.top <- .bottom[idx]
      gap.bottom <- .top[idx + 1]
      
      hz.template <- i[1, ids.top.bottom.idx]
[...]

PR #207, opened 2021-03-02
fillHzGaps: optimize RE: #205

brownag:


dylanbeaudette:
Great work, thanks for all of the faster code / ideas.

Notes to self / us before I forget: 
 * "grow" should really be reserved for extending existing horizons (aesthetic purposes typically)
 * filling gaps, and extending profiles (up/down) with empty horizons may all be possible with a single function:
   - fill up
   - fill down
   - fill gaps
   - all three

PR #211, opened 2021-03-03
estimateAWC and correctAWC using RO2 component lookuptable

brownag:


dylanbeaudette:
Thanks, this is great. We can hash-out the rest in #212.

PR #215, opened 2021-03-09
Remove aqp::union(), simplify combine() and overload c() S4-style

### LINKED ISSUE #161: Make aqp::union() robust to profile ID / site attr collisions.
```r
library(aqp)
library(soilDB)

x <- fetchOSD('musick')

x$hd <- hzDistinctnessCodeToOffset(x$distinctness, codes=c('very abrupt', 'abrupt', 'clear', 'gradual', 'diffuse'))
x$code <- 'OSD'

# hack: simulation with sd=0 -> replication of original data
s <- sim(x, n = 6, hz.sd = 0)

s$hd <- rep(NA, times=nrow(s))
l <- list()

hzdo <- c(0, 0.5, 2, 5, 15, 20)/2
for(i in seq_along(hzdo)) {
  ss <- s[i, ]
  ss$hd <- hzdo[i]
  l[[i]] <- ss
}

s <- aqp::union(l)
s$code <- c('none', 'very abrupt', 'abrupt', 'clear', 'gradual', 'diffuse')

## error on 2020-08-04
# https://github.com/ncss-tech/aqp/issues/161
z <- aqp::union(list(x, s))

## removing old ID "fixes" error:
# s$id <- NULL
# z <- aqp::union(list(x, s))

# no error when used like this
z <- aqp::union(list(s, x))

par(mar=c(0,0,1,0))
plotSPC(z, width=0.33, name.style='center-center', print.id=TRUE, hz.distinctness.offset = 'hd', label='code', hz.depths=TRUE, cex.names=0.8, plot.depth.axis=FALSE)
title('Horizon Boundary Distinctness', line=-1)
```

brownag:
- Removes `aqp::union` (previously deprecated after making it "safe" for SPCs)
- Adds `aqp::c(SPC)`
  - [x] <strike>call `base::c`</strike> use `unlist(..., recursive=FALSE)` for backwards compatibility when `x` is SPC but `...` is not all SPCs or have custom names
- Fixes #161


brownag:
> There is probably a slightly more elegant way to do this.
Specifically the {aqp} method should cleanly dispatch to `base::c()` if the inputs are not all SPCs

brownag:
Thanks @dylanbeaudette for probing me further on that, the fix to `plotMultipleSPC` shouldn't have been necessary, but helped me conceptualize what needed to happen.

https://github.com/ncss-tech/aqp/pull/215/commits/dc429a6bca2423ca44764dae9b3224c0e7cf267e should cover the cases where base-like behavior is needed (it uses `unlist(..., recursive=FALSE)` to first-order-simplify the non-SPC short circuit result)

dylanbeaudette (approved):
Thanks, looks good. Is the addition / overload of `c()` a convenience or to side-step namespace issues with `combine()`?

brownag:
That was the thought. If `combine` may continue to cause issues with deprecation in conjunction with dplyr, we can override base "combine" which is `c()`--"a generic function which combines its arguments."

Since we are making a method for the SPC `S4` the argument is different from the `c(...)` that most of us think about. It is a S4 generic, but with argument list `(x, ...)`. So, we make use of our S4 benefits, essentially, and overload c in an almost completely unambiguous way. The only issues I encountered were a couple rare cases where it was intended that `c(x=SPC,...=something_non_spc)` returned a regular list where the first element was `x` and was an SPC. That all should work as intended now, and result depends on what the ... is, essentially.

dylanbeaudette:
Great, thanks for the explanation. Merge it!

PR #221, opened 2021-04-20
Improved `parseMunsell`

pierreroudier:
This PR implements an improved `parseMunsell`, which now can handle less stable munsell notations, eg using other delimiters than `/`, or missing white spaces.

One study case remains to be implemented: `x.neutral <- parseMunsell('N 2/')`, but it should be possible to adapt this code, which essentially splits the munsell string into 3 parts.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aqp
 Version: 1.30
-Date: 2021-04-14
+Date: 2021-04-20
 Title: Algorithms for Quantitative Pedology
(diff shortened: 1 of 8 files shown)


pierreroudier:
Very close now. One test not passing: 

```
expect_silent( { d4 <- horizons(munsell2spc(sp3, .data = data.frame(foo = sp3$chip))) } )
```

dylanbeaudette:
> Very close now. One test not passing:
> 
> ```r
> expect_silent( { d4 <- horizons(munsell2spc(sp3, .data = data.frame(foo = sp3$chip))) } )
> ```

This works now that NA handling has been upgraded.

PR #224, opened 2021-05-18
Add subsetHz SPC method

brownag:
``` r
library(aqp, warn.conflicts = FALSE)
#> This is aqp 1.30
library(soilDB)

data("loafercreek")

# A horizons with top depth greater than 10cm
res <- subsetHz(loafercreek, grepl("A", hzname), hzdept > 10)

plot(res)
```


dylanbeaudette:
Great idea, this will simplify a lot of my redundant code.

PR #228, opened 2021-05-23
Fix estimateAWC errors on R 3.6.x

brownag:
- Uses `data.frame(stringsAsFactors=FALSE)` and a label column instead of `dput()` structure containing a factor column. 
- [x] Fixes factor levels and order / remove ambiguity resulting in base AWC mismatch (passes on R 3.6.x?)


brownag:
RE: rounding on <= R 3.6.x https://cran.r-project.org/web/packages/round/vignettes/Rounding.html

PR #237, opened 2021-08-27
Implement `required` argument for SPC metadata

brownag:
Implement `required` argument for SPC metadata: horizon designation nd texture class name

 - methods that call `hz*name()` metadata get/set method with the `required=TRUE` argument will error if not specified/invalid
 - consolidate docs
 - cleanup SPC S4 code/comments ~(more to do here)~

``` r
library(aqp)
#> This is aqp 1.31
data(sp3)
depths(sp3) <- id ~ top + bottom

# get the hz desgn
hzdesgnname(sp3)
#> [1] ""

# get the hz desgn (fail if empty)
hzdesgnname(sp3, required = TRUE)
#> Error: Horizon designation (Metadata element: 'aqp_hzdesgn') is not specified for this SoilProfileCollection. It is required.
#> See ??hzdesgnname

# set the hz desgn (fails; goo does not exist)
hzdesgnname(sp3) <- "foo"
#> Error: Horizon designation name (foo) not in horizon names.

# re-set the hzdesgn to "" (empty)
hzdesgnname(sp3) <- ""

# re-set the hzdesgn to "" (fails, because the resulting hz desgn is empty)
hzdesgnname(sp3, required = TRUE) <- ""
#> Error: Horizon designation name (Metadata element: 'aqp_hzdesgn') is not specified for this SoilProfileCollection. It is required.
#> See ??hzdesgnname

# ditto texcl
hztexclname(sp3)
#> [1] ""
hztexclname(sp3, required = TRUE)
#> Error: Horizon texture class (Metadata element: 'aqp_hztexcl') is not specified for this SoilProfileCollection. It is required.
#> See ??hztexclname
hztexclname(sp3) <- "foo"
#> Error: Horizon texture class name (foo) not in horizon names.
hztexclname(sp3) <- ""
hztexclname(sp3, required = TRUE) <- ""
#> Error: Horizon texture class name (Metadata element: 'aqp_hztexcl') is not specified for this SoilProfileCollection. It is required.
#> See ??hztexclname
```


dylanbeaudette:
I like this. Will check to see which functions I've contributed to in the past could benefit from this level of strictness.

PR #239, opened 2021-09-03
Add `as.character(<SoilProfileCollection>)`

brownag:
Currently, you can store a list of SoilProfileCollection as a column of a data.frame when protected with `I()`. This could be pretty convenient if you have multiple groups of pedons that you need to manage.

But if you try to print that data frame in the console, you will get an error because there is no way to represent the list of SPC as a character vector. 

If we were to define the S4 method `as.character()` for the SoilProfileCollection this would be resolved. 
This PR provides an example of how it could be defined; formatting could be changed as needed.

``` r
library(aqp, warn.conflicts = FALSE)
#> This is aqp 1.31

data(sp1)

data(sp2)
depths(sp2) <- id ~ top + bottom

test <- data.frame(id = 1:2, foo = I(list(sp1, sp2)))
test
#> Error in as.character.default(new("SoilProfileCollection", idcol = "id", : no method for coercing this S4 class to a vector

# define as.character(<SPC>)
setMethod('as.character', 'SoilProfileCollection', function(x, ...) {
  paste0('SPC<', length(x), '>')
})

# inspect after defining the as.character method
test
#>   id     foo
#> 1  1  SPC<9>
#> 2  2 SPC<18>

# use the list column
lapply(test$foo, function(x) profileApply(x, aqp::estimateSoilDepth))
#> [[1]]
#> P001 P002 P003 P004 P005 P006 P007 P008 P009 
#>   89   59   67   62   68  200  233  200  240 
#> 
#> [[2]]
#>  hon-1 hon-10 hon-11 hon-13 hon-14 hon-17 hon-18 hon-19  hon-2 hon-20 hon-21 
#>    170    225    220    301    320    295    300    410    130    283   1300 
#> hon-22  hon-3  hon-4  hon-5  hon-6  hon-7  hon-8 
#>   1300     78    147    180    184    370    390
```

### DIFF
--- a/R/Class-SoilProfileCollection.R
+++ b/R/Class-SoilProfileCollection.R
@@ -55,11 +55,6 @@ setClass(
   }
 )
 
-#' @export
-setMethod('as.character', 'SoilProfileCollection', function(x, ...) {
-  paste0('SPC<', length(x), '>')
-})
-
 # 2019-03-15: creating an empty SpatialPoints object requires more effort
 # c/o: https://gis.stackexchange.com/questions/291069/creating-empty-spatialpoints-or-spatialpointsdataframe-in-r
 # old: new('SpatialPoints')
@@ -420,7 +415,15 @@ setMethod(f = 'show',
 
           })
 
-
+#' @description `as.character()`: Character Representation of SoilProfileCollection Object
+#' @param x a SoilProfileCollection
+#' @param ... additional arguments (not used)
+#' @keywords internal
+#' @rdname show
+#' @export
+setMethod('as.character', 'SoilProfileCollection', function(x, ...) {
+  paste0('SPC<', length(x), '>')
+})
 
 #' @title Wrapper method for data.frame subclass conversion
 #'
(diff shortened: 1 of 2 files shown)


dylanbeaudette:
That is an interesting use case that I've never encountered before, neat. In retrospect, seems like a reasonable way to manage groups.

brownag:
I have been trying to carefully use list columns recently. I am not sure that this specific thing is the safest, but it seems to be possible. I suppose there are cases where things get converted to character might start "working" where previously they would have errored?

brownag:
And I guess to clarify my earlier statement, there is no way to represent the SPC as a vector in general, and as.character happens to be the default. I think the same thing could be achieved here with `as.vector(x)`

brownag:
Unless we can come up with a reason otherwise we will stick with the `"SPC<#>"` string format rather than something more verbose like `SoilProfileCollection(idname="foo", depths = c("top", "bottom"), ...)`. The former is concise and provides useful info.

We will add the number of horizons too. So it will be `SPC<number_of_sites, number_of_horizon>`

PR #247, opened 2021-12-14
"spatial predicates" for horizon data

brownag:
Here are some ideas I have been considering around indexing horizon data in a SoilProfileCollection based on "spatial predicates" that operate on the "horizon geometry".

I want to be able to find the horizons that are above or below arbitrary horizons for the whole collection.

**New function** `spc_horizonOffset()` allows for vertical offset (in terms of `[,j]` index units) of horizon IDs. 

Two aliases of this function are `horizon_above()` and `horizon_below()`; these use a different sign on the `offset` argument ("above" is negative, "below" is positive), and allow for `subset()`/`subsetHz()` like horizon-level filtering expressions in `...`

This is just an experiment with implementing an idea I've had for a while. I applied this new offset function to extend the `repairMissingHzDepths()` to be able to repair ALL missing bottom depths when top depths are present.


dylanbeaudette:
No rush, but wanted to give a heads-up on a pending CRAN release. Did you want this PR merged before that time?

brownag:
Should be good to go. Renamed to hzOffset/hzAbove/hzBelow for better parity with existing aqp horizon focused functions/naming style. Also updated docs for subset/subsetHz/hzAbove/hzBelow to give some warning about possible contexts where non-standard evaluation might not work as expected.

dylanbeaudette:
Thanks! That was fast. Will merge, and test with `dice()`.

PR #248, opened 2022-02-09
`slab()`: pass `formula` argument to aggregate.formula(x, ...)

brownag:
- adjusts for changes in R 4.2.x; `aggregate.formula()` now follows generic `aggregate(x, ...)` https://bugs.r-project.org/show_bug.cgi?id=18299
 - when passing a (named) list of arguments to `aggregate()` via `do.call()` need to _not_ specify argument name (which is `formula` R < 4.2, and `x` on R >= 4.2)
   - It is implied that `aggregate.formula()` will be called when first argument is class formula

### DIFF
--- a/R/slab.R
+++ b/R/slab.R
@@ -313,7 +313,7 @@
 
 	# optionally account for extra arguments
 	else {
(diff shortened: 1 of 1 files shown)


dylanbeaudette:
Thanks for noticing the errors on r-devel and suggesting a fix! This was totally off my radar.

PR #250, opened 2022-03-16
`spc2mpspline()`: upgrades

brownag:
Here are some optimizations to `spc2mpspline()` using recent upgrades to other functions that operate on whole SPCs. Even for small collections there is a performance improvement. Much less iteration is required to get the same result.

I would like to have a discussion on potentially relaxing some of the restrictions about depth of splining. In principle the splines do not need to be constrained to the shallowest bottom depth that has data for the whole collection. 

``` r
library(aqp)
#> This is aqp 1.42

data(sp1)

system.time(spc2mpspline(sp1, "prop"))
#> Loading required namespace: mpspline2
#>    user  system elapsed 
#>    0.23    0.02    0.25
system.time(aqp:::.spc2mpspline.old(sp1, "prop"))
#>    user  system elapsed 
#>    1.11    0.03    1.26
```


dylanbeaudette:
Nice! I'll take a look later tonight. TL/DR version:
  * yes: we should relax those constraints, I see no reason to limit to the most-shallow profile
  * I suspect that using `dice` will further improve run time

brownag:
> Nice! I'll take a look later tonight. TL/DR version:
> 
> * yes: we should relax those constraints, I see no reason to limit to the most-shallow profile
> * I suspect that using `dice` will further improve run time

OK, that is what I am now thinking. I don't have any workflows that specifically depend on that behavior, and I have a couple that involve calling spc2mpspline on single profile SPCs and recombining.

 
And the new function uses `dice()`! The old version used `slice()`.

brownag:
Updated in 3c45669f  

image
^This is the plot now generated at the end of test-spc2mpspline.R

R/spc2mpspline.R
```diff
@@ -35,9 +37,113 @@
 #'
 setMethod("spc2mpspline", signature(object = "SoilProfileCollection"),
           function(object, var_name = NULL,
-                           pattern = "R|Cr|Cd|qm",
-                           hzdesgn = guessHzDesgnName(object),
-                           ...) {
+                   pattern = "R|Cr|Cd|qm",
+                   hzdesgn = guessHzDesgnName(object),
```
dylanbeaudette, line 41: Since this is kind of important, I'd suggest following similar pattern matching of horizon designation in other functions and use what is set via or supplied. There are some cases where you don't have horizon designations (equal-interval sampling, or perhaps synthetic data) and would want to set this to NA or NULL to "turn off" this filtering.

R/spc2mpspline.R
```diff
@@ -35,9 +37,113 @@
...
+            hzbot <- horizonDepths(object)[2]
+            
+            # glom to "available interval" in each profile
+            spc.sub <- glom(object, object[[hztop]][object[, 1, .HZID]], 
+                                    object[[hzbot]][object[, , .LAST, .HZID]])
+            
+            # debug : inspect  horizon values for var_name
+            #plot(spc.sub[1:10,], color=var_name)
+            
+            # only take profiles that have 100% data coverage in above interval
```
dylanbeaudette, line 67: Yes, I don't think there is any way around this constraint.

R/spc2mpspline.R
```diff
@@ -35,9 +37,113 @@
...
+            # calculate the deepest top depth and shallowest bottom depth
+            mindepth <- max(spc.sub[, 1][[hztop]], na.rm = TRUE)
+            
+            # TODO: decide if it is necessary to be so restrictive
+            maxdepth <- max(c(minDepthOf(spc.sub, pattern = pattern, hzdesgn = hzdesgn)[[hztop]], 
+                            max(spc.sub)), na.rm = TRUE)
+            
+            # we will only make interpolations that the "whole SPC supports"
+            # the thought is that these 1cm slices will be further aggregated downstream
+            spc.sub <- trunc(spc.sub, mindepth, maxdepth)
```
dylanbeaudette, line 81: This is one of the constraints we were thinking about relaxing, yes?

dylanbeaudette (commented):
Looks good to me other than a couple of questions.

R/spc2mpspline.R
pierreroudier: I agree, ms-splines are a major tool for DSM-type approaches,  where it's very, very common to have zero hz designations.

dylanbeaudette (approved):

brownag:
I did some more improvements here. Now mpspline2 does essentially all the handling of missing data--further reducing input requirements. Also added `method` argument to allow output spline values at original depth support (`est_icm`) or static depth intervals (`est_dcm`) specified via `d` argument to `mpspline()` in addition to the default of `est_1cm` 1cm output.

``` r
library(aqp) 
#> This is aqp 1.42
data(sp1, package = "aqp")

res0 <- spc2mpspline(sp1, 'prop', 
                     hzdesgn = 'name',
                     d = c(0, 5, 15, 30, 60, 100, 200, 300))
#> Loading required namespace: mpspline2
plot(res0, color = "prop_spline", divide.hz = FALSE)
```



``` r
res1 <- spc2mpspline(sp1, 'prop', 
                     method = "est_icm", 
                     hzdesgn = 'name',
                     d = c(0, 5, 15, 30, 60, 100, 200, 300))
plot(res1, color = "prop_spline")
```



``` r
res2 <- spc2mpspline(sp1, 'prop', 
                     method = "est_dcm", 
                     hzdesgn = 'name',
                     d = c(0, 5, 15, 30, 60, 100, 200, 300))
plot(res2, color = "prop_spline")
```

dylanbeaudette:
Nice. This is much closer to the results / interpretation of how I expected such a function to work with an SPC. I'm curious about the differences between the original values vs. EA spline estimates when using the source horizon depths. For example is the '-10' in the color legend a result of a tiny (-0.001) or large (-2) undershoot of the spline? I'll have to look over the original paper + code--I don't recall what constraints are in place to prevent the EA spline estimates from exceeding plausible limits.

brownag:
> I'm curious about the differences between the original values vs. EA spline estimates when using the source horizon depths. For example is the '-10' in the color legend a result of a tiny (-0.001) or large (-2) undershoot of the spline?

It is a tiny undershoot. I also thought it is weird. I considered truncating these slightly negative values to 0 via an argument.

 > I'll have to look over the original paper + code--I don't recall what constraints are in place to prevent the EA spline estimates from exceeding plausible limits.

So far I have only seen it when `method = "est_icm"`

brownag:
Merging this in for further testing. Definitely worth looking into the numerical anomalies with `est_icm` and looking for datasets that might break the function. Future upgrades might allow for interpolating multiple properties in one call.

PR #257, opened 2022-08-17
Update intro vignette

brownag:
This brings the vignette (based on http://ncss-tech.github.io/AQP/aqp/aqp-intro.html) up to date, fixes some inconsistent formatting, re-orders a few sections, and adds content to stub sections.

### DIFF
--- a/vignettes/Introduction-to-SoilProfileCollection-Objects.Rmd
+++ b/vignettes/Introduction-to-SoilProfileCollection-Objects.Rmd
@@ -1,6 +1,9 @@
 ---
 title: "Introduction to SoilProfileCollection Objects"
-output: rmarkdown::html_vignette
+output: 
+  html_document:
+    toc: true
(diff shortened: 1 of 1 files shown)


dylanbeaudette:
Thanks. At some point, I need to retire/redirect the other version here.

brownag:
Yeah, I would say (and the approach I have taken with stats class materials that have moved) use a JavaScript redirect in the HTML file of "old" one to move to the "new" one, that way old link references go to the new material.

Use something like this in http://ncss-tech.github.io/AQP/aqp/aqp-intro.html
```html
<!-- ... -->
<script>
  window.location.replace("http://ncss-tech.github.io/aqp/articles/Introduction-to-SoilProfileCollection-Objects.html");
</script>
<!-- ... -->
```

Could also use the CRAN equivalent (once 1.43+ is submitted)

Issue #229, opened 2021-07-13
implement `weights` argument for `slab()`


brownag:
A planned feature for `slab()` is to (re-)support site-level weights. See #3

When specifying some dummy weights the following error message is issued
``` r
library(aqp)
#> This is aqp 1.30
#> 
#> Attaching package: 'aqp'
#> The following object is masked from 'package:stats':
#> 
#>     filter
data(sp1)
depths(sp1) <- id ~ top + bottom
slab(sp1, fm = ~ prop, weights = 1:length(sp1))
#> Error: weighted aggregation is not yet supported
```

Example use case: 
 - component percent weighted averages from an SPC derived from one or more SSURGO mapunits worth of horizon data (containing multiple soil components). 
 
 See https://github.com/ncss-tech/soilDB/pull/194 for some sample DMU record IDs for weighted average aggregation and validation against Web Soil Survey aggregation

dylanbeaudette:
Thanks, this is described in #3.

brownag:
Note to consider option of supplying weights as a character vector length 1 column name, or as a vector equal in length to number of profiles.
In principle I think it would be possible to support a matrix of weights, and/or horizon level weights.

dylanbeaudette:
> Note to consider option of supplying weights as a character vector length 1 column name (implemented in #258), or as a vector equal in length to number of profiles. In principle I think it would be possible to support a matrix of weights, and/or horizon level weights.

Yes: the original implementation (2008?) allowed for site-level or horizon-level weights. It wasn't entirely clear to me at the time how to implement horizon-level weighting into things like weighted SD over intervals > 1cm thickness.

I'd prefer that the updates as part of this issue and related PR implement site-level weighted first, via named column from site. Once the details of horizon-level weighting are determined, then I'd prefer to use a similar approach via named column from horizons.

brownag:
> I'd prefer that the updates as part of this issue and related PR implement site-level weighted first, via named column from site. 

This is done

Issue #115, opened 2020-01-21
`data.table` re-write of `slice()`


dylanbeaudette:
## A Plan
 1. finish  `dice` 
 2. test / document `dice`
 3. deprecate `slice`
 4. fix many affected functions / documentation  (note `slice` max + 1 convention)
 5. enable slicing shortcut through the SPC `k-index` (maybe)

## TODO
 - [x] finish tests
 - [ ] side/side comparisons with `slice` in all aqp functions (this will take time, and aqp 2.0 is not backwards compatible)
 - [x] sometimes `fm = 0:z ~ .` results in `z+1` slices (bug related to filling to bottom of z-index, z +1)
 - [x] NA-padding to depth via `fillHzGaps` (default = FALSE), important for `slab` and `profile_compare`
    - [x] fully-integrate newly optimized `fillHzGaps`, consider separate arguments to control filling / padding
 - [x] formula interface (backwards compatibility, less data to push around):
    - [x] slice selection via integer vector
    - [x] variable selection via NSE `var1 + var2 + ...`
    - [x] ERROR when asking for slices entirely within horizon gaps or invalid horizons (addressed in #205, pending optimization)
 - [x] .pctMissing eval (new function)
    - [x] optimize with `data.table` (see warnings...)
    - [x] DT warnings related `:=` (see below)
 - [x] ERRORS on NA bottom depth (consider `repairMissingHzDepths()`)
 - [x] horizon depth logic errors: flag (`@metadata`) / subset:
    - [x] horizon level `byhz = TRUE` (should eventually do gap-filling with NA records #205)
    - [x] profile level `byhz = FALSE`
 - [ ] ~~integrate into `slab()` (problems as of 2022-01-04), or integrate into slab-rewrite (better).~~ (different PR / issue)

## Long-Term
 - [ ] future optimization `setindex()` vs. `setkey()` <-- this sorts the data (probably not worth the extra effort)
 - [ ] overlapping (A/E type) horizons (#88)

## Updates
 * 2021-02-11: just about ready for use. This is not a direct replacement for `slice`, maybe introduce now and phase-in by aqp 2.0.
 * 2022-01-04: much more robust, direct replacement for `slice` in most cases. will attempt to change all `slice` → `dice` in the SPC introduction and see what happens. a messages is now printed when using `slice`, will be deprecated in aqp 2.0.

[...]

PR #258, opened 2022-08-18
Updates to `dice()`, `slab()`, `genSlabLabels()`

brownag:
These are some updates that resolve some tasks from #115, and implements #229

Notably, `slab()` now uses `dice()` so users will not get confused when they get a deprecation message RE: `slice()` while using `slab()`. I have had several questions from users about this (essentially: how do I use dice() with slab()?)

```r
library(aqp)
#> This is aqp 1.43
data(sp1, package = 'aqp')
a.max <- slab(sp1, fm = ~ prop)
```

This PR has attempted to cover / ensure that the following are addressed:
 - issues related to the  `fm = 0:z ~ .` results in `z+1` slices (fixed insofar as they affect `slab()`, I don't think it is fully fixed on `dice()` side yet)
 - handled 0 thickness horizons in `slab()`, whether or not `strict=TRUE`
 - fixed mismatching lengths of slab labels v.s rows in data
 - added and exported `slab_function()` a helper method for accessing the internal slab methods provided by the package; now used in method definitions rather than internal functions (could use more documentation on arguments/patterns for custom functions, and ensuring extra arguments all works)
 - added a weighted variant of the numeric/quantile method via `Hmisc::wtd.quantile()`, and support for custom weighted functions (such as `weighted.mean()`)
 
I am sure this has not yet caught everything, but I spent some time ensuring all the existing tests worked, and added some more that stress a few other features, as well as new tests for the now-implemented `weights` argument. It all appears to be functioning as expected in the test cases, though I bet there are still edge cases or un-tested use cases that still have issues.

### DIFF
--- a/R/dice.R
+++ b/R/dice.R
@@ -27,6 +27,8 @@
 #' 
 #' @param byhz Evaluate horizon depth logic at the horizon level (`TRUE`) or profile level (`FALSE`). Invalid depth logic invokes `HzDepthLogicSubset` which removes offending profiles or horizon records.
 #' 
+#' @param verbose Print information about object size/memory usage. Default: `FALSE`
+#' 
 #' @details For large and potentially messy collections that include missing horizon bottom depths, or 0-thickness horions, consider using `repairMissingHzDepths()` before `dice()`.
 #' 
 #'
--- a/R/dice.R
+++ b/R/dice.R
@@ -36,19 +36,26 @@
 #' 
 #' @export
 #' 
-dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, strict = TRUE, byhz = FALSE) {
-  
+dice <-  function(x,
+                  fm = NULL,
+                  SPC = TRUE,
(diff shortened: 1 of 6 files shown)


brownag:
One more note: I added a stub for a "weighted categorical/factor" aggregation function, but it currently does nothing. I think I might need to spend a little more time looking at the `cpm`/"class probability mode" logic before I get into thinking about NA handling and weighting for categories.

dylanbeaudette:
I haven't finished an in-depth review, but before this gets too complex I'd like to split the `dice/slab` changes into two pieces. I don't know how to work the git magic to do that, will need your help on that. This is a complex tangle. Also, I think that these changes (finally) to slice, slab, dice, etc. warrant a 2.0 release.

brownag:
So you are asking for two pull requests, one for dice and one for slab/genSlabLabels? 

PR has been split into two draft PRs. The `slab()` PR branches from / depends on the changes in `dice()`:

 - https://github.com/ncss-tech/aqp/pull/260
- https://github.com/ncss-tech/aqp/pull/259

PR #259, opened 2022-08-19
Updates to `slab()`

brownag:
These are some updates that close https://github.com/ncss-tech/aqp/issues/229

 - [x] `slab()` now uses `dice()` so users will not get confused when they get a deprecation message RE: slice() while using slab(). I have had several questions from users about this (essentially: how do I use dice() with slab()?)

```r
library(aqp)
#> This is aqp 1.43
data(sp1, package = 'aqp')
a.max <- slab(sp1, fm = ~ prop)
```

- [x] fixes mismatching lengths of slab labels v.s rows in data; allows for overlap; see `.genSlabLabels2()`
- [x] added and exported `slab_function()` a helper method for accessing the internal slab methods provided by the package; now used in method definitions rather than internal functions 
  - [x] add more documentation on arguments/patterns for custom functions, and ensuring extra arguments (or lack thereof)) work robustly
- [x] added a weighted variant of the numeric/quantile method via `Hmisc::wtd.quantile()`, and support for custom weighted functions (such as `weighted.mean()`)

-  weighted variant of categorical method (created stub function `.slab.fun.factor.weighted()`) now in #269


R/genSlabLabels.R
```diff
@@ -1,54 +1,70 @@
-## TODO: documentation / generalization
-# note source data must be normalized via dice() first, e.g. all share the same number of horizons
+# does not assume logical horizons (may overlap, resulting in different # of slices per profile)
+.genSlabLabels2 <- function(spc, diced, slab.structure = 1) {
```
dylanbeaudette, line 2: Some commentary + pretty formatting would help me understand what is going on here.

R/segment.R
```diff
@@ -3,19 +3,20 @@
...
+#'
 #' @param object either a `SoilProfileCollection` or `data.frame`
 #' @param intervals a vector of integers over which to slice the horizon data (e.g. `c(25, 100)` or `25:100`)
-#' @param trim logical, when `TRUE` horizons in `object` are truncated to the min/max specified in `intervals`. When `FALSE`, those horizons overlapping an interval are marked as such. Care should be taken when specifying more than one depth interval and `trim = FALSE`.
+#' @param trim logical, when `TRUE` horizons in `object` are truncated to the min/max specified in `intervals`. When `FALSE`, those horizons overlapping an interval are marked as such. Care should be taken when specifying more than one depth interval and \code{`trim = FALSE`}.
```
dylanbeaudette, line 10: why the `\code{}` ?

tests/testthat/test-slab.R
```diff
@@ -35,30 +33,131 @@ test_that("basic slab functionality", {
   expect_true(any(grepl('p.q', nm)))
 })
 
+test_that("slab structure for a single slab", {
```
dylanbeaudette, line 36: Thanks for the extra tests.

R/slab.R
```diff
@@ -1,20 +1,16 @@
-
-
 # default slab function for categorical variables
 # returns a named vector of results
 # this type of function is compatible with aggregate()
-.slab.fun.factor.default <- function(values, cpm) {
+.slab.fun.factor.default <- function(values, cpm, ...) {
```
dylanbeaudette, line 4: \dots used for optional weights?

R/slab.R
```diff
@@ -27,202 +23,163 @@
...
 
 	return(pt)
-	}
-
-
-## TODO: make these exported functions with documentation (https://github.com/ncss-tech/aqp/issues/99)
-
+}
 
+.slab.fun.factor.weighted <- function(values, w, na.rm = TRUE, ...) {
```
dylanbeaudette, line 28: I'm thinking that weighted / unweighted functions be merged into a single function with a NULL `w` argument. When it is specified the weighted version of e.g. `mean` is used.

R/slab.R
```diff
@@ -27,202 +23,163 @@
...
 	# get extra arguments: length of 0 if no extra arguments
 	extra.args <- list(...)
 
-	## important: change the default behavior of data.frame and melt
-	opt.original <- options(stringsAsFactors = FALSE)
-
 	# get unique list of names in object
 	object.names <- unique(unlist(names(object)))
 
-	# get number of profiles
```
dylanbeaudette, line 110: I had to look this up, but removing these lines has to do with swapping out the previous `genSlabLabels`. Some bread crumbs would be helpful next time.

R/slab.R
```diff
@@ -236,127 +193,135 @@
...
+	if (!missing(weights)) {
+	  if (missing(slab.fun)) {
+	    # default _weighted_ aggregation fun is .slab.fun.numeric.weighted
+	    FUN <- .slab.fun.numeric.weighted
+	    # custom weighted aggregation fun should take first argument as values 
+	    # second argument is "weights" argument
+	  } else {
+	    FUN <- slab.fun
+	  }
+	  wt <- eval(weights)
```
dylanbeaudette, line 239: What does this do?  Required for DT syntax?

R/slab.R
```diff
@@ -236,127 +193,135 @@
...
+	  d.slabbed <- as.data.frame(d.long[, as.data.frame(t(FUN(value, .SD[[wt]], ...))), by = c('variable', g, 'seg.label')])
```
dylanbeaudette, line 240: Please expand a little bit for clarity, whitespace / newlines.

R/slab.R
```diff
@@ -236,127 +193,135 @@
...
+	  
+	  # reset factor levels in d.long[[value]]
+	  if (.factorFlag) {
+	    d.long[['value']] <- factor(d.long[['value']], levels = original.levels)
+	  }
+	  
+	  # process chunks according to group -> variable -> segment
+	  # NA values are not explicitly dropped
+	  if (length(extra.args) == 0) {
+	    d.slabbed <- aggregate(aggregate.fm, data = d.long, na.action = na.pass, FUN = slab.fun)
```
dylanbeaudette, line 260: How hard would it be to convert `aggregate` → DT approach? We can do this in a follow-up PR to make testing simpler. This conversion is complicated by the possibility of numeric vs. factor data.

dylanbeaudette (commented):
Looks good, please have a look at specific comments. I would like to do some testing before fully converting the unweighted stuff to DT.

R/slab.R
brownag: All of the slab functions now have `...` in their definition for consistency and possible future expansion

R/slab.R
brownag: This allows for programmatically specifying the weights argument, rather than as a constant. Required to play nice with data.table non-standard evaluation. This line of code has further been updated because using the name `wt` within the SPC as the weights column could also cause problems with NSE.

R/slab.R
dylanbeaudette: OK. I'm fine with full replacement of `aggregate` in favor of DT. Either two calls or one that can generalize across weighted / un-weighted based on passing arguments. This should greatly improve the performance.

R/slab.R
brownag: A comment to indicate that I removed (now unused) variables? This is probably something I caught with the RStudio heuristics. I will try to remember to summarize such things in PR content

R/genSlabLabels.R
brownag: I added documentation and some comments. I am not sure if we want to export this function if this is the one we are going to use internally? I found that some of the assumptions in the old genSlabLabels didnt generalize as well to the less logical inputs, so wanted to create something that did the same thing / had apprx same output, but since inputs are different I created new function/definition

R/slab.R
brownag: This probably can be done along with the below comments about `aggregate()` -> data.table

R/slab.R
brownag: OK, in 2da417ad have it working with _three_ data.table calls. Will need to do a little thinking and possibly refactoring of the slab functions to get it down to a single call supporting weighted/unweighted/factor/numeric value input.

dylanbeaudette: Three calls is simpler to debug / test. Lets leave it alone for now. This is looking good. I can't do much more testing / review until tomorrow. Thanks.

dylanbeaudette:
I thought of one potential issue, found last night while working on `NCSP` and `dice`: when using a formula that specifies all variables `~ .` the horizon metadata variables will be included in the calculation of percent missing. This isn't usually a problem, but something to keep in mind as we move forward. `slab` always includes a named list of variables on the formula RHS, so this isn't a problem. We just need to be consistent in how we carry forward "horizon metadata", as a future plan for `slab` includes output as an SPC.

brownag:
> I thought of one potential issue, found last night while working on `NCSP` and `dice`: when using a formula that specifies all variables `~ .` the horizon metadata variables will be included in the calculation of percent missing. This isn't usually a problem, but something to keep in mind as we move forward.

It's funny you mention this, since given the discussion we had on `dice()` with that type of formula I considered this in testing

> `slab` always includes a named list of variables on the formula RHS, so this isn't a problem. 

Yep. I think it would be cool if `slab()` could support the `~ .` syntax. 

> We just need to be consistent in how we carry forward "horizon metadata", as a future plan for `slab` includes output as an SPC.

Yeah, agreed. Would like to see a method for converting slab output to SPC, I actually was going to draft one... and if we were to set that up... then we of course would need to address the metadata issue for consistency in "SPC methods that return SPCs"

brownag:
Assuming I navigated the parallel changes to master branch with the new "factor" check dataset correctly-- I find no difference between the v1 and v2 slab methods when using `~ genhz` formula.

image

brownag:
I've looked over the KSSL check code and made updates last weekend, the numeric discrepancy for that check set is in the Musick profiles that don't pass a logic check.

Remaining comments and questions:
 - Are the "changes" in value  (generally ~rounding error) and contributing fraction acceptable? These differences only occur in the case of profiles that have invalid depth logic to begin with.
 - Is there any issue with contributing fraction calculation in new code? Are we properly accounting for overlaps and gaps that are now allowed as inputs?

brownag:
From my testing I find that the cpm=2 argument also gives the same results for ~genhz formula between version 1 and 2.

PR #260, opened 2022-08-19
Updates to `dice()`

brownag:
Minor changes to `dice()`  to support its use in `slab()`


R/dice.R
```diff
@@ -163,21 +172,31 @@ dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, str
...
   # safe wrapper to base::seq
   # NA in from/to (hz depths) not allowed
-  # TODO: from == to (0-thickness) horizons ??
   .seq.safe <- function(from, to, by) {
-    if(any(is.na(from)) | any(is.na(to))) {
+    if (any(is.na(from)) | any(is.na(to))) {
       stop('corrupt horizonation, consider using `strict = TRUE`', call. = FALSE)
     }
-    
+    if (from > to) {
```
dylanbeaudette, line 179: Can you please comment on what this is doing?

R/dice.R
```diff
@@ -163,21 +172,31 @@ dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, str
...
+      toold <- to
+      to <- from
+      from <- toold
+    }
     return(
       seq(from = from, to = to, by = by)
     )
   }
   
+  # from == to (0-thickness) horizons -- these should be removed -AGB 2022/08/17
```
dylanbeaudette, line 189: How does this interact with `strict == TRUE`? I see the utility, but this is new behavior.

R/dice.R
```diff
@@ -188,30 +207,21 @@ dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, str
...
+    times = h.sub[[htb[2]]] - h.sub[[htb[1]]]
   )
   
-  
   # assemble slice LUT for JOIN
   s <- data.table(
     sliceID = sliceIDs, 
-    .sliceTop = tops, 
-    .sliceBottom = bottoms
+    .sliceTop = tops[seq_along(sliceIDs)],
```
dylanbeaudette, line 217: Short description of why the new indexing?

dylanbeaudette (commented):
Thanks for looking at this. I've got a couple of questions as inline comments. Will do some testing of the branch locally later today.

R/dice.R
brownag: This removes 0 thickness even when `strict=FALSE`, they should already be removed when `strict=TRUE`

brownag: My thought is that its not really possible to have a 0 thickness slice--it has no "volume"--I could be off on this reasoning though.

R/dice.R
brownag: `mapply()` and `.seq.safe()` were returning non-conforming vectors. Will review this to make sure it is still needed

R/dice.R
brownag: I think it was sufficient to drop the 0 thickness horizons. The inversion issue had to do with sequences from 1 to 0 I believe

R/dice.R
brownag: This also was not needed, as filtering 0 thickness horizons is sufficient

brownag:
Thanks for your comments.

I had several pieces that do not seem to be necessary when the 0 thickness horizons are filtered out ahead of time (regardless of whether `strict` is TRUE. PR has been updated and marked as ready for review. 
 
I _think_ that it doesn't make sense to return a record for a 0-thickness slice --from a practical standpoint such a slice has no associated weight.

brownag:
OK, I have now incorporated changes that are make it possible to support overlapping horizons (if desired). The related changes in slab are not working (yet), but I think this makes the needed changes to `dice()`

I believe what is still in need of review is the default behavior of `strict`/`byhz`

R/dice.R
dylanbeaudette: Agreed, that 0-thickness hz contains no useful information. We can return to this after dealing with other issues. Since this cleaning step will likely transcend `strict = TRUE`, we may want to move it "up" in the order of operations.

R/dice.R
```diff
@@ -147,76 +155,44 @@ dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, str
...
   }
   
-  tops <- mapply(
-    FUN = .seq.safe, 
-    from = h[[htb[1]]], 
-    to = h[[htb[2]]] - 1, 
-    by = 1, 
-    SIMPLIFY = FALSE
-  )
+  h.sub <- as.data.table(h[which(!zthk),])
```
dylanbeaudette, line 168: `h` should already be a data.table object, why the re-casting?

R/dice.R
```diff
@@ -147,76 +155,44 @@ dice <- function(x, fm = NULL, SPC = TRUE, pctMissing = FALSE, fill = FALSE, str
...
   
-  tops <- unlist(tops)
+  # expand 1 unit slices to max depth of each profile
```
dylanbeaudette, line 170: Is this more efficient than the previous version? If it is, then I'd like to expand it out a little bit for readability.

R/dice.R
brownag: Indeed this is not needed, it may have been a result of debugging with a data.frame at some point. Can remove

R/dice.R
dylanbeaudette: I'm all for it, maybe we can look at it together so I can better understand exactly what is happening in there. I'll just expand out a little for visual clarity. Robust > simplicity here.

dylanbeaudette:
I've found a new issue with the new `dice` as implemented in this PR, related to discrete slices specified in a formula interface. I think that this has to do with slice vector creation, as it doesn't happen in `slice` and I am pretty sure did not happen before this PR. I'll check on that next.

Here is a quick example from `misc/sandbox/dice.R`, with helper functions.
```r

par(mar = c(1, 1, 3, 1))

# plot original + diced sketches
.sideBySidePlot <- function(d, s, ...) {
  
  # adjust IDs on slices
  profile_id(s) <- sprintf("%s*", profile_id(s))
  g <- combine(d, s)
  
  # original + sliced, same color scheme
  plotSPC(g, name = NA, divide.hz = FALSE, width = 0.3, default.color = 'grey', ...)
}


# check that slices contain the same data as source
.slicesAreSame <- function(d, s) {
  ## double check data are conserved
  .s <- horizons(s)[, c('hzID', 'p1', horizonDepths(d))]
  .d <- horizons(d)[, c('hzID', 'p1', horizonDepths(d))]
  .m <- merge(.s, .d, by = 'hzID')
  
  # remove NA, slices within gaps / below profile max depth
  .m <- na.omit(.m)
  
  res <- all(.m$p1.x == .m$p1.y)
  if (!res) {
    return(.m[which(.m$p1.x != .m$p1.y), ])
  } else {
    return(res)
  }
  
}


set.seed(1010)
d <- lapply(as.character(1:10), random_profile, n = c(6, 7, 8), n_prop = 5, method = 'LPP', SPC = FALSE)
d <- do.call('rbind', d)
depths(d) <- id ~ top + bottom

## continuous slices
s <- dice(d)
.sideBySidePlot(d, s, color = 'p1')
.slicesAreSame(d, s)


## TODO: figure out issues with discrete slices

## discrete slices

# works as expected
.slices <- c(5)
s <- dice(d, fm = .slices ~ .)
.sideBySidePlot(d, s, color = 'p1')
.slicesAreSame(d, s)

# sliced data do not entirely math original data
.slices <- c(150)
s <- dice(d, fm = .slices ~ .)
.sideBySidePlot(d, s, color = 'p1')
.slicesAreSame(d, s)

# sliced data do not entirely math original data
.slices <- c(5, 10, 15, 25, 100, 190)
s <- dice(d, fm = .slices ~ .)
.sideBySidePlot(d, s, color = 'p1')
.slicesAreSame(d, s)

```

tests/testthat/test-dice.R
```diff
@@ -84,10 +81,15 @@ test_that("formula interface", {
   )
   
   expect_true(all(profileApply(s4, nrow) <= 31))
+})
   
+test_that("formula interface (with non-standard evaluation)", {
+  # NSE (these work interactively, but not in testthat env)
+  .slices <-  0:30
```
dylanbeaudette, line 88: Is this just remind us about NSE / testthat?

brownag: Well, I suppose it is a reminder of the syntax in general, as it _can_ be used interactively. 
Mainly was just showing that the loss of ability to use NSE in this context is not a regression from slice->dice. 

We can comment it out / get rid of it so we don't have unnecessary number of tests of "expected" behavior

And when slice is deprecated or if removed from namespace would need modification anyway.

dylanbeaudette: cool, thanks

dylanbeaudette:
I'm done testing this PR, and can confirm that all of the functions that rely on `slice()` have been successfully ported over to `dice()` and tested. Looks good to merge.

brownag:
Alright! I will now merge this and have rebased the slab changes onto it.

PR #261, opened 2022-08-24
Add pH / reaction class conversion functions

brownag:
I wrote these functions a few months back to use in a report that summarizes SSURGO component data values/ranges.

Since we have tools for converting between textural classes and their numeric counterparts, I thought it might be useful to have an analogue for pH in aqp.

Open to any suggestions to improve/rename/add arguments etc. if this is something we might be interested in including

TODOs:

 - [x] Proper handling of NA
 - [x] add a small function for setting reaction class factor levels
 - [x] optionally set factor levels in pH → reaction class conversion, via argument/toggle
 - [x] create the source table as a saved .rda, as opposed to read.table() on each function call

EDIT: in response to TODO items I have:

 - Added function `ReactionClassLevels()` which is used in `ph_to_rxnclass()` via `as.is=FALSE` argument
 - Added internal function .reactionclass() with data.frame definition, and .reactionclass2() that accesses reactionclass.rda; the reaction classes in the dataset are documented as coming from SSM (2017)


dylanbeaudette:
Great idea, there are many applications for these kind of conversions. It would be nice to use this functionality in SKB when parsing OSDs or similar narratives. `.parse_pH_class()` attempts to do this, but it doesn't do anything except find strings and set reasonable factor levels.

A couple of suggestions:
 1. add a small function for setting reaction class factor levels
 2. optionally set factor levels in pH → reaction class conversion, via argument/toggle
 3. create the source table as a saved .rda, as opposed to `read.table()` on each function call

brownag:
Added to TODO list., thanks @dylanbeaudette!

brownag:
For what it is worth RE: "create the source table as a saved .rda, as opposed to read.table() on each function call" it is about 5x more efficient to simply define a function that uses the `read.table()` syntax rather than reading from .rda file stored in the package data folder.

This would possibly change if we switched over all of the public-facing datasets to work with LazyData: true, but that would require significant refactoring of other data that are loaded.

dylanbeaudette:
Using the `data.frame` approach seems like a good idea. My main issue with reading raw text via `read.table` was transcription / editing errors.

dylanbeaudette:
Also, when would someone use the `halfclass` argument?

brownag:
> Also, when would someone use the `halfclass` argument?

It is actually the entire reason that this function was created. Because when going from numeric -> class it is possible to split classes to get additional detail on the range. It is particularly helpful for the neutral class, but relevant to all relatively wide classes.

brownag:
Despite my reservations about the high likelihood of analysis errors with this approach... I have updated these new functions to default to returning ordered factors with dropped levels. I feel that this is at least consistent with existing similar functions in the package (e.g. texture classes)

brownag:
Despite my reservations about the high likelihood of analysis errors with this approach... I have updated these new functions to default to returning ordered factors with dropped levels. I feel that this is at least consistent with existing similar functions in the package (e.g. texture classes)

dylanbeaudette:
> Despite my reservations about the high likelihood of analysis errors with this approach... I have updated these new functions to default to returning ordered factors with dropped levels. I feel that this is at least consistent with existing similar functions in the package (e.g. texture classes)

There are two distinct issues here: ordered factors and dropping levels. I'd suggest `drop = FALSE` as a default. Kind of like `SoilTextureLevels()`. This function is a little different as it doesn't really do anything. I'd also like to ask Stephen about `ssc_to_texcl()` and setting `drop = FALSE` for the default.

brownag:
> I'd suggest `drop = FALSE` as a default. Kind of like `SoilTextureLevels()`. 

`ReactionClassLevels()` doesn't even provide the option to drop the levels--it would make no sense to do that. My concern is about `ph_to_rxnclass()`, which as you recognize, takes data with an intrinsic order (and indeed, specific codes [possibly]] related to that order) as input. It wouldn't make sense to drop the levels from either of the Levels functions as they take no data as input, but it is a concern for the conversion functions. The idea that you could get the same numeric value for different levels from different datasets is my primary concern.

> I'd also like to ask Stephen about `ssc_to_texcl()` and setting `drop = FALSE` for the default.

I think we probably don't need `droplevels` arguments to any of the functions. It is fewer characters to type to just do `droplevels(somefunction(somedata))` rather than `somefunction(somedata, droplevels = TRUE))`. That said, I don't think this issue is setting out to make a precedent here.

> That said, it is your call. 
I don't really think it is, I think we are obligated to at least try and make these things consistent, so that is what I have done. If we change our mind on how factors are handled (we should) then we can deal with that later.

dylanbeaudette:
> > I'd suggest `drop = FALSE` as a default. Kind of like `SoilTextureLevels()`.
> 
> `ReactionClassLevels()` doesn't even provide the option to drop the levels--it would make no sense to do that. My concern is about `ph_to_rxnclass()`, which as you recognize, takes data with an intrinsic order (and indeed, specific codes [possibly]] related to that order) as input. It wouldn't make sense to drop the levels from either of the Levels functions as they take no data as input, but it is a concern for the conversion functions. The idea that you could get the same numeric value for different levels from different datasets is my primary concern.

I was only referring to the conversion functions as needed / using a `drop` argument. 

> > I'd also like to ask Stephen about `ssc_to_texcl()` and setting `drop = FALSE` for the default.
> 
> I think we probably don't need `droplevels` arguments to any of the functions. It is fewer characters to type to just do `droplevels(somefunction(somedata))` rather than `somefunction(somedata, droplevels = TRUE))`. That said, I don't think this issue is setting out to make a precedent here.

Fine by me for simple cases where the function does "one thing", these conversion functions are a good example. For something more complex like `fetchNASIS()` *dropping levels* is far more nuanced and important--thus an argument.

> > That said, it is your call.
> 
> I don't really think it is, I think we are obligated to at least try and make these things consistent, so that is what I have done. If we change our mind on how factors are handled (we should) then we can deal with that later.

Since this is aqp, we don't have to maintain perfect consistency with soilDB although that is a fine goal. I'm fine with merging this now, either with / without `drop` argument.

brownag:
> Since this is aqp, we don't have to maintain perfect consistency with soilDB although that is a fine goal. 

Agreed. I am mostly talking here about `ssc_to_texcl()` which is the analogous function to `ph_to_rxnclass()`, both of which are in aqp.

> I'm fine with merging this now, either with / without `drop` argument.

Great. I just made the levels all lowercase (again for consistency). `ph_to_rxnclass()` includes the `droplevels` argument just like `ssc_to_texcl()`

PR #264, opened 2022-09-22
Horizon metadata columns

### LINKED ISSUE #263: Always retain metadata, even with explicit subsetting as in dice()
Should functions that operate return SPC objects always return full metadata such as `hzdesgnname`? I have no preference either way, but preservation seems like a good strategy no matter the implied subsetting.

brownag:
Provides a mechanism `hzMetadata()` and internal method `.hzMetadataNames()` to close #263 

Applied to (so far)
 - `dice()`
 - `show(<SoilProfileCollection>)`


dylanbeaudette:
Easy, I wanted to review this a little longer.

R/dice.R
```diff
@@ -199,7 +199,8 @@ setGeneric("dice", function(x,
   names(s)[1] <- hzidn
   
   # FULL JOIN via fast data.table compiled code
-  res <- merge(h, s, by = hzidn, all = TRUE, sort = FALSE)
+  hm <- merge(h, hzMetadata(x), by = c(idn, hzidn), all.x = TRUE, sort = FALSE)
```
dylanbeaudette, line 202: This isn't going to work in general: the metadata need to be included in `vars` further up, otherwise there will be duplication. This happens when e.g. including the horizon designation in the RHS of the formula or specifying all hz variables with `.`.

```r
set.seed(1010)
d <- lapply(as.character(1:10), random_profile, n = c(6, 7, 8), n_prop = 5, method = 'LPP', SPC = FALSE)
d <- do.call('rbind', d)
depths(d) <- id ~ top + bottom

hzdesgnname(d) <- 'name'

# duplication
dice(d, fm = 25 ~ p1 + name, SPC = FALSE)

# OK
dice(d, fm = 25 ~ p1, SPC = FALSE)

# duplication
dice(d, fm = 25 ~ ., SPC = FALSE)
```

R/Class-SoilProfileCollection.R
```diff
@@ -295,19 +295,9 @@ setMethod(f = 'show',
...
-              # registered in a slot
-              idx <- match(c(idname(object),
-                             hzidname(object),
-                             horizonDepths(object),
-                             hzdesgnname(object)), names(h))
-            } else {
-              # undefined
-              idx <- match(c(idname(object), hzidname(object), horizonDepths(object)), names(h))
-            }
+            # move IDs and depths, horizon designation/texture if available
```
dylanbeaudette, line 298: This is much cleaner, I like it.

dylanbeaudette (commented):
The metadata getting function makes a lot of sense, but I'd prefer either always returning the depths or logical flag to toggle. Also, use of the metadata is not a drop-in replacement for the existing `dice` code.

dylanbeaudette:
One more question: why return a copy of the data, when most of the time we are interested in the column names? For example `horizonDepths`, `horizonNames`, `idname`, `hzidname`. In most cases, especially in `dice`, all of the relevant subsetting is by column name.

R/Class-SoilProfileCollection.R
```diff
@@ -1147,6 +1121,33 @@ setMethod("horizonNames", signature(object = "SoilProfileCollection"),
             return(res)
           })
 
+.hzMetadataNames <- function(object, depths = character(0), ...) {
```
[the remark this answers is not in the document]
brownag: Sure, I think I had `horizonDepths(object)` as the default when i first drafted this function, then changed the default to "no depths". Agreed, makes sense and should be fixed in https://github.com/ncss-tech/aqp/commit/8d22184e78c51db5e69379dd85ac50cd4dd9f132.

R/dice.R
brownag: You are right, I think this is fixed now with https://github.com/ncss-tech/aqp/commit/721ef8a435e020d28d948583901a72f7cd600ca0

brownag:
> One more question: why return a copy of the data, when most of the time we are interested in the column names? 
You are right, we only need the internal method for use in dice() which already uses its own copy of horizon data. Should be fixed in https://github.com/ncss-tech/aqp/commit/721ef8a435e020d28d948583901a72f7cd600ca0

PR #265, opened 2022-09-28
vectorize `perturb()`

brownag:
PR to close #233

PR #268, opened 2022-10-05
`generalize.hz()` ideas

brownag:
This is a draft PR containing some ideas from a few months ago to improve/extend utility of `generalize.hz()` function and integrate it better in other aspects of the SPC.

 - Adds S4 method (`generalizeHz()`) for SPC _and_ character. This _could_ allow us to transition to the "camel case" style of functions like this rather than the "." name style (though this isn't needed)
 - Override to let ordered factors be created without specifying `hzdepm` (possibly not super useful in current form, but there is room for additional helper functions to work with this)
 - Introduces concept of the "gendesgnname" 
 
Possible further extension of the "gendesgnname" could be adding related metadata functions (getters/setters). Recent work related to the preservation of metadata would lower the barrier to adding a new "horizon metadata column"

From my view, it is valuable to have a function that takes an SPC and returns an SPC that does the "genhz labeling". I don't feel strongly that this needs to be incorporated, but the branch was open and had fallen behind, so might be worth a discussion at least.


dylanbeaudette:
Thanks. Good ideas, I'll give my short comments here and we can discuss later in the week if you have time.

> This is a draft PR containing some ideas from a few months ago to improve/extend utility of `generalize.hz()` function and integrate it better in other aspects of the SPC.
> 
> * Adds S4 method (`generalizeHz()`) for SPC _and_ character. This _could_ allow us to transition to the "camel case" style of functions like this rather than the "." name style (though this isn't needed)

That is fine, no reason to use `.` in function names if there is a reasonable alternative. I can see why it might be useful in some cases to use the pattern `SPC <- f(SPC)`, perhaps with pipes, it is not as flexible. There are many cases where myself and others have created several candidate GHL columns and assignment to new hz level attributes is very simple. S4 methods make it possible to have both interfaces to the function--so I'm supportive of a new SPC-specific method.

> * Override to let ordered factors be created without specifying `hzdepm` (possibly not super useful in current form, but there is room for additional helper functions to work with this)

I agree, the order is implied by `new` and `pattern`. That is how I've used it 90% of the time. Honestly, I can't think of a reason to use the hz midpoint right now.

> * Introduces concept of the "gendesgnname"

Good idea, it can only help with adoption / use of this concept. I'd suggest `GHL` based on prior usage and simplicity.

> 
> Possible further extension of the "gendesgnname" could be adding related metadata functions (getters/setters). Recent work related to the preservation of metadata would lower the barrier to adding a new "horizon metadata column"

Sure. Again, GHL seems much simpler. I have noticed that moving metadata *through* functions like `dice` can be lead to new bugs. One example I discovered in `NCSP` when the `removed.profiles` metadata was set prior to running `NCSP`.

> From my view, it is valuable to have a function that takes an SPC and returns an SPC that does the "genhz labeling". I don't feel strongly that this needs to be incorporated, but the branch was open and had fallen behind, so might be worth a discussion at least.

Now is a good time to make big changes.

brownag:
> That is fine, no reason to use `.` in function names if there is a reasonable alternative. I can see why it might be useful in some cases to use the pattern `SPC <- f(SPC)`, perhaps with pipes, it is not as flexible. There are many cases where myself and others have created several candidate GHL columns and assignment to new hz level attributes is very simple.

Strong agree, which is why I created `generalizeHz(<character>)`... just a wrapper around the existing `generalize.hz()`. 

I would count myself among the "others" who have used several "GHL"-style columns at once. I have found it very useful in my soil taxonomy lab+morphology related evaluations for the standards committees.

> I agree, the order is implied by `new` and `pattern`. That is how I've used it 90% of the time. Honestly, I can't think of a reason to use the hz midpoint right now.

I think that the midpoints can be helpful when the regex gets really hairy. i.e. where there is logic embedded in the sequence the patterns are applied because later patterns can overwrite earlier ones and the order from `new`/`pattern` is _not_ a depth-based sequence. I have examples of this. Granted, these casescould likely be refactored such that they do work in order. As far as "opportunities" I was thinking we probably could promote existing approaches for estimating a reasonable order (i.e. `guessGenHzLevels()`) for a given SPC.

> > * Introduces concept of the "gendesgnname"
> 
> Good idea, it can only help with adoption / use of this concept. I'd suggest `GHL` based on prior usage and simplicity.
> Possible further extension of the "gendesgnname" could be adding related metadata functions (getters/setters). Recent work related to the preservation of metadata would lower the barrier to adding a new "horizon metadata column" ... GHL seems much simpler. 

Fine by me. The metadata entry could be `metadata(spc)$ghl` (or `ghlname`) and the getter/setter could be `GHL()` and `GHL<-()`.

> I have noticed that moving metadata _through_ functions like `dice` can be lead to new bugs. One example I discovered in `NCSP` when the `removed.profiles` metadata was set prior to running `NCSP`.

I saw your changes related to that. I would think someone might want to know if their `NCSP()` result was derived from a subset of the set of profiles they started with (i.e. what they had before `dice()`/`HzDepthLogicSubset()` which gets called _inside_ `NCSP()`)... but I can see why we might not want to persist all metadata in all cases, and there is definitely no need to do so if it does not make sense with the result.

For functions that take SPC as input and produce an SPC we may need to recognize that some of them are "derived" and some of them are "new" SPCs--despite always being based in some way on the source SPC. 

To draw a line in the sand somewhat arbitrarily, to me anything that uses `aqp::combine()` internally is a "new" SPC. This is notable because the order, number of sites/horizons etc. will almost always change in this case. With recent perturb updates (#265) now are back to only one function that uses the SoilProfileCollection constructor (depths<-)... which is of course the that "new" SPCs are made.

> Now is a good time to make big changes.

Hence all the noise from me in the issue tracker :)

dylanbeaudette:
All good on merging this after a 2.0 release, or did you want to get it in before?

brownag:
It would not take much work to define the get/set `GHL()` methods as we have discussed here. Can do that before release. Were you planning on releasing today?

brownag:
OK, marked ready for review. Take a look at https://github.com/ncss-tech/aqp/pull/268/commits/e7628c6aee647d01d024ae2d6304fb3785bb91aa to see how I anticipate GHL being implemented. If we would like to not "require" GHL be set for some of these functions we can change the defaults and add some extra handling code for previous hard coded/default arguments for column names i.e. "genhz"/"name"

PR #277, opened 2023-02-01
accumulateDepths: remove constraint requiring profiles with >1 horizon

brownag:
I wanted to be able to re-accumulate depths from zero for _portions_ of profiles... such as those created by `subsetHz()`... and I noticed/was reminded of constraint on minimum number of horizons > 1 in `accumulateDepths()`. This was not explicitly documented, nor was there a very good reason for it in the code.

In this PR the constraint has been removed. Now the following works without the message `"horizons (n=2) dropped from set"` and the removal of two of the profiles.

``` r
library(aqp)
#> This is aqp 2.0
data("loafercreek", package="soilDB")
twobee <- subsetHz(loafercreek, grepl("2BC*t", hzname))
plot(twobee)
```

``` r
plot(accumulateDepths(twobee, hzdatum = -twobee[,,.FIRST]$hzdept))
```

<sup>Created on 2023-02-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/accumulateDepths.R
+++ b/R/accumulateDepths.R
@@ -210,18 +210,17 @@ accumulateDepths <- function(x,
   dat$thk <- dat[[.BOTTOM]] - dat[[.TOP]]
   dat$thk[is.na(dat$thk)] <- 0
   
-  # need more than 1 record
-  test1 <- dat[, .N, by = .BYLIST1]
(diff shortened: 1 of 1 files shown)


dylanbeaudette:
Cool. I don't recall what the original constraint was for either.

PR #280, opened 2023-02-26
Fixing tests and issues with minimal install

brownag:
Some notes/fixes from testing on a fresh machine with a minimal set of dependencies.
 - removes {scales} dependency from `plotSPC()`
 - removes {plyr} usage from test-SPC-combine.R
 - <strike>skip `mixMunsell()` tests if {gower} not installed</strike>
   fixed bad logic that required {gower} for `mixMunsell(..., mixingMethod="exact")`
 - avoid usage of `aqp:::` in tests
 - skip `profile_compare()` (deprecated) tests if {scales} is not installed
 - skip some `slab()` tests if {Hmisc} is not installed

### DIFF
--- a/R/mixMunsell.R
+++ b/R/mixMunsell.R
@@ -218,18 +218,18 @@ mixMunsell <- function(x, w = rep(1, times = length(x)) / length(x), mixingMetho
   mixingMethod <- match.arg(mixingMethod)
   
   # multiple matches only possible when using mixingMethod == 'reference'
-  if((n > 1) & mixingMethod != 'reference') {
+  if((n > 1) && mixingMethod != 'reference') {
     stop('`n` is only valid for `mixingMethod = "reference"`', call. = FALSE)
   }
   
   # mixed spectra and multiple matches only possible when using mixingMethod == 'reference'
-  if(keepMixedSpec & ! mixingMethod %in% c('reference', 'exact')) {
+  if(keepMixedSpec && !mixingMethod %in% c('reference', 'exact')) {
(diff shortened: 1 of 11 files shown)


dylanbeaudette:
Are we 100% sure that the `scales` replacement is 100% compatible with all current use cases? I haven't had a chance to test that. Will have more time this week.

brownag:
Every instance I came up with gave pretty much identical results. Can you think of something that is not currently tested that might have broken?

brownag:
Also--just to clarify--scales is still used conditionally if it is available-- to "test" you need to remove {scales} from your R lib

dylanbeaudette:
> Also--just to clarify--scales is still used conditionally if it is available-- to "test" you need to remove {scales} from your R lib

Ah, ok. I didn't realize that--I thought that this was a replacement. I'll do some side/side testing and report back in the related issue. I've been wanting to remove scales from imports but haven't found the time or motivation to finish that. It works well and is less annoying than the old plyr code.

Thanks for the recent cleanup.

brownag:
So, scales I believe is still used for the (now deprecated) profile_compare() code. 

I believe the new function is a suitable replacement in so far as it works with tests and the color mapping appears to be identical in the cases we test. However I did not really rigorously test it other than trying some of the different data type inputs (numeric, factor/character, ordered factor) I did a bit of side by side and saw no obvious differences, but perhaps there is something I have missed in my replacement code. In theory I think scales could be dropped from Suggests entirely, but I didnt want to do that in this PR.

As a side motivation for this conditional usage of scales is I believe it means I can also drop scales from suggests from soilDB and remove a lot of the boilerplate related to scales for any examples that use scales in both packages

PR #283, opened 2023-03-28
`sieve()` for classifying arbitrary coarse fractions

brownag:
This is following up from a soilDB issue  https://github.com/ncss-tech/soilDB/issues/217 where we wanted to make the workhorse sieve function (with custom thresholds) available for more general usage outside soilDB.

The function defined by this PR should be adequate as is to replace the internal soilDB `.sieve()` method (sometime in the future, probably after adding a aqp >2.0 import req). In the meantime we can do whatever we want to improve what this one does, and then adjust soilDB accordingly.


dylanbeaudette:
Nice, thanks for implementing. I like the more generalized approach in this version. I've made a couple of tiny adjustments. A tiny change, but what do you think about "class_" as the generic name vs. "grp" ?

brownag:
> A tiny change, but what do you think about "class_" as the generic name vs. "grp" ?

Totally fine with that! I pondered the default value there for about 0.5 seconds after realizing the code didnt work if sieves wasn't named; if you prefer "class_" as default for result names when `sieves` is an unnamed vector, that sounds great

brownag:
Thanks also for the additional example! 

Also for docs probably this also needs a reference to the SSM, field book, or similar for the (default) size classes.

dylanbeaudette:
OK, thanks. Good ideas. I'll adjust default prefix, and a citation.

Soil Science Division Staff. 2017. Soil survey manual. C. Ditzler, K. Scheffe, and H.C. Monger (eds.). USDA Handbook 18. Government Printing Office, Washington, D.C.

brownag:
A couple other things I noticed:
 - "fine gravel" is one of 3 gravel subclasses and possibly shouldn't be a default cut in aqp
   - soilDB can use non-default `sieves`, and soilDB methods are smart enough to make sure the simplified data are composited correctly (i.e. fine gravel are included in "total gravel" volumes)
 - "parafragments" should probably be "pararock fragments"; can also mention artifacts

dylanbeaudette:
Another idea, but I don't have a use case: optionally returned ordered factors.

brownag:
> Another idea, but I don't have a use case: optionally returned ordered factors.

This briefly occurred to me too; possibly a function like `SoilTextureLevels()` to generate the `sieves` argument instead of current usage of if() inside function def? This would provide an easy mechanism to swap between different standard systems used in US and abroad

dylanbeaudette:
> > Another idea, but I don't have a use case: optionally returned ordered factors.
> 
> This briefly occurred to me too; possibly a function like `SoilTextureLevels()` to generate the `sieves` argument instead of current usage of if() inside function def? This would provide an easy mechanism to swap between different standard systems used in US and abroad

I like that idea. Now ~90% implemented via `fragmentClasses()`

dylanbeaudette:
Have a peek, I'm done here and happy with the results.

PR #284, opened 2023-04-01
Control section of Histosols and Histels

### LINKED ISSUE #133: control section of Histosols and Histels
Evaluating criteria for the "arbitrary control section" depths used in the Histosols and Histels. 

This relies on code being used in the new branch ncss-tech/aqp@mollic

```
[snip]
```

This calculates theoretical surface tier thickness not truncated to RLL

```
surface.tier.thickness <- profileApply(r02_histosols, histosol.surface.tier)
```

4 out of 5 RO2 Histosols (81%) have the 0-30cm surface tier (i.e. do not have fibric in 0-60cm)
NB: even without checking sphagnum content or bulk density, we can see this case is not common in RO2.

```
> prop.table(table(surface.tier.thickness))
surface.tier.thickness
       30        60 
0.8178439 0.1821561 
 
```

```
# Calculate control section top and bottom depths
#  these are truncated to RLL where appropriate, and use the surface tier thickness calculated
hcs.bounds <- profileApply(r02_histosols, histosol.control.section, frameify = TRUE)

sum(hcs.bounds$hcs_b > 130) / nrow(hcs.bounds)
sum(hcs.bounds$hcs_b == 160) / nrow(hcs.bounds)
```

- Approximately 7.5% of RO2 histosols have a CS bottom depth >130cm
- Approximately 3% of RO2 histosols CS descriptions actually extend to 160cm

brownag:
will close #133 

Items to resolve:
 - [x] How to handle the actual "mineral" PSCS component for families of Terric subgroups? Possibly this is more detailed than the control section key/function needs to cover
 - [x] Add examples that use the O-horizon-related rules and Organic specific limits (i.e "W" and "ff" horizon designations)


brownag:
> Did you want this in 2.0?

If it is finished in time then it would be fine to include it, but I wouldn't say it is necessary to include. These soils (Histosols, Histels) are already minimally handled by a prior patch that is already on master branch.

dylanbeaudette:
ok, no rush. but I'm going to try for this week.

PR #294, opened 2023-08-16
Fix examples for `slab()`

brownag:
@dylanbeaudette found a couple issues with dontrun examples in `slab()` that I have made some preliminary corrections for.

Issue 1: related to the data type of an all NA column being logical rather than double -- corrected by using `as.numeric()` in `.slab.fun.numeric.HD()`

Issue 2: related to a case we didn't have tests for and apparently had not been checked/identified as a problem since the `slab()` rewrite 
 -  `.genSlabLabels2()` has been corrected to adjust for minimum slice depth within a slab. Previously was returning all NA labels in cases where `slab.structure[1] > 0`. Added a new test for this case. 

Also small correction to `slab()` example for recent update to `get.ml.hz()` in https://github.com/ncss-tech/aqp/commit/a8932cfed95caedc0f418c56a43e0747b4632fcc

### DIFF
--- a/R/slab-utils.R
+++ b/R/slab-utils.R
@@ -96,23 +96,23 @@ genSlabLabels <- function(slab.structure = 1, max.d = NULL, n.profiles = NULL, s
   
   # calculate a sequence of cumulative depths and corresponding indices for each slab
   j <- diff(i)
+  hzd <- horizonDepths(spc)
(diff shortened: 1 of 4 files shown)


dylanbeaudette (approved):
Thanks for the quick changes. Looks good to me.

PR #303, opened 2023-11-14
Move {sp} to suggests

### LINKED ISSUE #289: sp evolution
Currently w/ {sp} >=1.6.x the following message is issued on load of aqp (and soilDB, or anything else that imports aqp)

```
The legacy packages maptools, rgdal, and rgeos, underpinning this package
will retire shortly. Please refer to R-spatial evolution reports on
https://r-spatial.org/r/2023/05/15/evolution4.html for details.
This package is now running under evolution status 0 
```

We still import the `SpatialPoints` object and import/export the `coordinates()` method for SoilProfileCollection, which invokes this dependency on {sp}. 

All {sp}-based code has been refactored in aqp 2.0 such that there is no reliance on {sp} internals for function results. Current CRAN version (aqp 1.42) is compatible with the sf-based {sp} backend (i.e. "evolution status" `2`) and does not require use of rgdal/rgeos.

At this point the SpatialPoints slot of the SPC is unused (superseded by new column metadata identifying geometry columns), and `coordinates()` `proj4string()` and their replace methods are deprecated (replaced by `initSpatial<-` & `getSpatial()`, `prj<-` and `prj()`). We could consider planning changes to the SPC object definition for aqp 2.1.x. The SpatialPoints slot is an "implementation detail" so explicit access of spc@sp has not been encouraged (though it may be used in the wild, as there was no direct accessor for it). Removing the unused `@sp` slot would be a breaking change to the S4 definition.

My understanding is that this month the evolution status will be updated from `0` to `2` (https://r-spatial.org/r/2023/05/15/evolution4.html). This is expected to continue to  have no effect on aqp itself, but users will get the message on load even after this change. Currently the message is issued even when manually setting `options(sp_evolution_status=2)` before load.

I am creating this issue primarily to identify that these messages from {sp} are a "known" issue in the {aqp} dependency tree; users may be confused by the message but at this point please be assured aqp is consistent with sp evolution process, though still reliant on some sp class definitions for backward compatibility purposes.

brownag:
Draft PR that removes the `@sp` slot from the SoilProfileCollection definition, and generally prevents loading of {sp} when {aqp} is loaded.

This breaking change is planned for aqp 2.1.x, will close #289


dylanbeaudette:
Good idea, thanks for getting this moving. Now is a good time to break from {sp}.

dylanbeaudette:
Almost ready for this. aqp 2.0.2 is on CRAN now which resolved a number of unrelated issues that I needed to clear out first.

brownag:
This is ready for review, and can be merged _after_ CRAN release of 2.0.3

brownag:
It appears the "missing data" vignette build is failing due to hzdesgnname not being set for jacobs2000. Note to check/rebuild the rdas without `@sp` slot and make sure relevant metadata are set before merge

PR #308, opened 2024-02-24
`flagOverlappingHz()` performance improvements

brownag:
`flagOverlappingHz()` is called by `fillHzGaps()` which is used by several higher level functions including:
- `dice()` / `slab()` 
- `profileInformationIndex()`
- `spc2mpspline()`

The current implementation of `flagOverlappingHz()` on master branch is slow, based on iteration over single profile SPCs, and complex operations on each element... it is the major bottleneck in `fillHzGaps()`.

This PR improves processing time by approximately 4 orders of magnitude with a 10,000 profile SoilProfileCollection. It also allows for flagging overlap in horizons that have missing depths (a TODO item), improves tests, and ensures that _imperfectly_ overlapping horizons are not flagged.

```r
library(aqp)
#> This is aqp 2.0.3

x <- data.table::rbindlist(lapply(1:10000, random_profile, SPC = FALSE))
depths(x) <- id ~ top + bottom

# flag overlap in 10k random profiles (that have no overlap)
system.time(flagOverlappingHz(x))

## BEFORE
#>   user  system elapsed 
#> 98.309   0.483  98.877 

## AFTER
#>  user  system elapsed 
#> 0.002   0.006   0.013 
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -15,6 +15,6 @@ URL: https://github.com/ncss-tech/aqp, https://ncss-tech.github.io/AQP/
 BugReports: https://github.com/ncss-tech/aqp/issues
 Language: en-US
 Encoding: UTF-8
-RoxygenNote: 7.3.0
+RoxygenNote: 7.3.1
 Roxygen: list(markdown = TRUE)
 VignetteBuilder: knitr
--- a/R/flagOverlappingHz.R
+++ b/R/flagOverlappingHz.R
@@ -34,27 +34,26 @@
 #' 
 flagOverlappingHz <- function(x) {
   h <- horizons(x)
+  idn <- idname(x)
   hzd <- horizonDepths(x)
   
-  # extract horizon depths
-  .tops <- h[[hzd[1]]]
-  .bottoms <- h[[hzd[2]]]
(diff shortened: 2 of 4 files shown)


brownag:
On second thought, iteration may be necessary here (i.e. https://github.com/ncss-tech/aqp/pull/308/commits/8f5e02185caab75648aeb09cfc4cd93e3ac70464 may need to be reverted/altered), or a few more checks and tests for the multiprofile case to be properly covered. Will add a test to this that ensures logic is corrected.

dylanbeaudette:
Nice improvements and clever approach to NA depths. Thanks for working on this.

brownag:
> Nice improvements and clever approach to NA depths. Thanks for working on this.

Thanks!

I encountered it trying `fillHzGaps()` on ~13k KSSL pedons via `fetchLDM()` and was surprised how long it took, then realized that there were some notes on need to optimize it. I guess I have never noticed issues with smaller SPCs, but is pretty apparent when working with the LDM snapshot.

dylanbeaudette:
Agreed, that bottleneck had to go. Thanks for tackling it.

brownag:
Is this OK to merge now?

PR #317, opened 2024-06-24
Fix compatibility with R <4.1/4.2

brownag:
This PR fixes some syntax in recent additions for compatibility with R versions less than 4.2.

Currently usage of the base pipe `|>` (R 4.1) and underscore placeholder `_` (R 4.2) in code and examples leads to parsing issues that prevent installing/loading/checking packages on earlier versions of R. Our current minimum R supported version is 3.5. We run continuous integration checks on R 3.6.

Recently a few suggested packages related to documentation/testing (testthat, rmarkdown, knitr, evaluate) have been bumped to R 4.0 minimum to conform with new "tidyverse" guidelines. These packages can be safely skipped, as they only pertain to certain development-related use cases. Eventually we may need to increase minimum R required version when required (imported) packages increase, but for now we can safely install and check on R < 4.0 by ignoring these packages in the CI.

From my view we should probably support R 3.6 for a while longer, as it is still out there in the wild on platforms that are not as easily updated. For instance many folks may have containers or HPC systems with access to just standard repositories. As an example R 3.6.3 is the default `r-base` package available on Ubuntu 20.04 LTS, which is still supported for the time being (about another year).


brownag:
w/  https://github.com/ncss-tech/aqp/pull/317/commits/d35e601eb36f299dac667a2e79d4c79ae5505207 CI for R 3.6 skips dependencies on tidyverse packages that require R >=4.0

dylanbeaudette (approved):
My only suggestion is to consider using extra newlines with (>2x) nested function calls, as a compromise. For example:

```r
lapply(
split(), 
function(i) {
...
}
)
```
@smroecker : do you have any external tests to double-check?

brownag:
> My only suggestion is to consider using extra newlines with (>2x) nested function calls, as a compromise. For example:
I cleaned up some of the syntax and whitespace here: https://github.com/ncss-tech/aqp/pull/317/commits/085d4bdd049fa8301c153d1ec1dbf87b6c47113b

Also fixed a couple deprecation messages that said old function names "will be" deprecated--if we use `.Deprecated()` they _are_ deprecated already.

dylanbeaudette:
Perfect, thanks!

smroecker:
Looks good. Thanks for making the changes Andrew!

Issue #318, opened 2024-09-26
munsell2rgb: color value of 8.5 and 9.5


brownag:
I noticed when working with the KSSL pedon morphologic companion database recently that there are some non-standard color warnings issued: `non-standard notation in Munsell value, use getClosestMunsellChip()`

I was curious if these were obsolete values in the database, but it turns out that `colorvalue` of 8.5 and 9.5 are legitimate choices (on the "white" page). Unlike some chromas (0, 5, 7) the half chip color values are not obsolete in the NASIS domains. 

Probably these can be handled by interpolation similar to how we handled `colorvalue` 2.5 (i.e https://github.com/ncss-tech/aqp/issues/178)

There are 56 instances of this in the morphologic database corresponding to KSSL pedons (latest dated 2021-09-14), and 560 total in full NASIS pedon snapshot (latest dated 2024-09-06)

``` r
library(soilDB)
dbQueryNASIS(NASIS(dsn = "D:/Geodata/soils/NASIS_Morphological_09142021/NASIS_Morphological_09142021.sqlite"),
             "SELECT * FROM phcolor") |> 
  uncode() |> 
  subset(colorvalue %in% c("8.5", "9.5"))
#> Loading required namespace: RSQLite
#>        OBJECTID phiidref seqnum colorpct colorhue colorvalue colorchroma
#> 2846       2846  6640453      1       NA     10YR        8.5           0
#> 2889       2889  6640454      1       NA     10YR        8.5           1
#> 214507   214507  6641422     NA       NA    7.5YR        8.5           3
#> 214834   214834  6640474     NA       70    7.5YR        8.5           2
#> 214835   214835  6640475     NA       70    7.5YR        8.5           2
#> 214836   214836  6640476     NA       70    7.5YR        8.5           2
#> 214837   214837  6641411     NA       25    7.5YR        8.5           2
#> 214838   214838  6641414     NA       75    7.5YR        8.5           2
#> 215742   215742  6640485     NA       70    7.5YR        8.5           1
#> 215743   215743  6641426     NA       NA    7.5YR        8.5           1
#> 215744   215744  6641427     NA       NA    7.5YR        8.5           1
#> 215745   215745  6641428     NA       NA    7.5YR        8.5           1
#> 216454   216454  6641428     NA       NA    7.5YR        8.5           1
#> 266846   266846   986282     NA       NA     10YR        8.5           0
#> 266847   266847   986283     NA       NA     10YR        8.5           0
[...]

PR #319, opened 2024-09-26
Add half-value Munsell chips beyond 2.5

dylanbeaudette:
Starting from #318, I've added all half-value Munsell chips to the `munsell` LUT used in aqp. This works well, but breaks existing tests, and many years of expected behavior. A result of `10YR 3.5/4` isn't always the best answer, even though it might be the closest conversion from e.g. sRGB coordinates or the result of a simulated mixture.

At this point, I think that it makes sense to just interpolate 2.5, 8.5, 9.5 value chips + spectra to minimize complexity in related functions and current behavior. A more intensive addition to the LUT and reference spectra should be another issue/PR.

TODO:
  - [ ] ~~Think about new arguments to color conversion functions the limit results to "standard" value chips~~
  - [x] finish interpolation of half-value spectra
  - [x] update all tests, many will fail at first
  - [x] more tests on expanded `munsell` LUT
  - [x] update documentation
  - [x] re-run closest munsell chips code -> update associated .rda

Also worth including:
  - [x] update rounding logic in `munsell2rgb()`
  - [ ] ~~implement optimization in `munsell2rgb()`~~
  - [x] misc. TODOs in `munsell2rgb()`
  - [x] double-check `getClosestMunsellChip()` behavior


dylanbeaudette:
Interpolation of Munsell reference spectra results in some odd outliers, at the edges of reference spectra. Will deal with that in a different issue.

dylanbeaudette:
I think this is just about done, now with a smaller scope.

@brown-ag at your convenience, can you please try re-building the closest munsell chips data and update the associated test? Those tests are currently skipped. Thanks.

brownag:
Should be done in https://github.com/ncss-tech/aqp/pull/319/commits/f24b51bd3d6a12788c8c48ea4af8f1d06bcec2ac

dylanbeaudette:
I've noticed some subtle differences in simulated mixtures of Munsell colors:

Before: 2.5GY 5/1
image


After: 2.5GY 4/2
{E89B6F3B-0385-4963-B905-02C7929F9C1A}

misc/utils/Munsell/local-functions.R
```diff
@@ -38,8 +38,47 @@ interpolateChroma <- function(m.i) {
...
+interpolateValue2 <- function(m.i) {
+  
+  # can only proceed with >=2 rows
+  # some combinations of hue, value, chroma have 1 row.
+  # there will be other combinations created by split() with 0 rows
+  if(nrow(m.i) < 2) {
+    return(NULL)
+  }
+  
+  # linear interpolation ~ munsell value
```
brownag: Update comment

brownag:
> I've noticed some subtle differences in simulated mixtures of Munsell colors:

I am not sure I completely understand why adding 8.5 and 9.5 value chips would change this result. Is this because of the use of `splinefun()` for interpolating values v.s. former linear interpolation via `approxfun()`?

dylanbeaudette:
Adding the 8.5 and 9.5 chroma entries to the `munsell` LUT and reference spectra required a re-write of multiple interpolation steps. In doing so I updated:
  * converted all interpolation to `splinefun`, vs. use of `approxfun` in the interpolation of a single, new 2.5 value chip/spectra
  * new interpolation of value/chroma/spectra now use the entire set of available points vs. just those bracketing the "missing" points → this leads to a more stable curve
  * added a check/fix for spectra reflectance < 0, caused by spline interpolation overshoots at a few extremes. This requires a little more testing as part of a new issue to be created. Previously this was only done for chroma, I added a similar check/fix for value
 
In short, the "addition" of 8.5 and 9.5 value chips / spectra triggered a cascade of updates and enhancements.

dylanbeaudette:
OK. I think that we are good to go here.

PR #337, opened 2025-12-17
convert stringr to stringi to base R

dylanbeaudette:
Slightly lower dependency stack, #331.


dylanbeaudette:
I agree that this can all be done in base R. 

For the time being, I think that the low dependency count + syntactic clarity of stringi is worth it. I plan to revisit further down-sizing of dependencies in the future; for example, I would like to adapt/integrate functions called from the colorspace package.

dylanbeaudette:
The point of this PR was to move away from stringr, after thinking more on the original discussion.

Syntactic clarity is for present and future me; both of whom enjoy using stringi functions to accomplish current and planned tasks. I agree that nearly everything can be done with base R. I'm sure we could also do everything with binary and bit-shifting operations.

Now, I hadn't realized that compiling stringi on my GFE takes 8.5 minutes and involves downloading 12 MB of source. That is a really long time. However, compiling sf and related packages is also a slog.

I'll review the affected files again and make a decision.

brownag:
We did discuss the compilation time and size of the library. I am sorry I did not go back to your issue and add more context, but I thought that was clear and agreed upon.

>  compiling sf and related packages is also a slog.

This is why it is in Suggests. It is only needed for the specific case of casting the   site data of an SPC to an sf data.frame. If a user does not need this functionality, they can readily do an imports only installation and ignore it. At this time, the same cannot be said for stringi or stringr.

PR #341, opened 2026-01-18
feat: add custom help handler for `$` .DollarNames autocompletion

brownag:
This implementation of `.DollarNames.SoilProfileCollection` S3 method supersedes #340
 - Exports .DollarNames.SoilProfileCollection as an S3 method
 - Adds custom help handler for interactive use in RStudio generation of help tooltip and additional information

The trick to getting the tooltips to work in RStudio (RE: #340) is setting a custom `helpHandler` attribute on the .DollarNames result that points to a function. This bypasses the typical process used for inspection of objects for generating help text. 
.DollarNames alone (without the help function attribute) is enough to get autocompletion in the vanilla R console or Rgui.

I studied how this was implemented for `reticulate` `python.builtin.object` to figure out how to do it. What reticulate is doing is a lot more complicated (it is parsing python docs into these structures), I was able to abstract out the relevant parts of the machinery for use on the SPC. This new RStudio-specific SoilProfileCollection help handler is implemented as the internal function `.aqp.rs.rpc.spc.HelpHandler` and is only attached in interactive RStudio sessions. This behavior can be toggled off using `options(.aqp.rs.rpc.integration=FALSE)`.  In theory, similar patterns could be implemented for other IDEs.

Now, provided the aqp package has been loaded into the namespace, when `$` autocompletion triggers we get a SPC specific output that contains not only the column name dropdown, but also custom tooltip has slot information, and structure (class, length, first several values), as seen below:

image

When you press F1, the `$,SoilProfileCollection-method` documentation opens.

### DIFF
--- a/R/SoilProfileCollection-operators.R
+++ b/R/SoilProfileCollection-operators.R
@@ -485,9 +485,13 @@ setReplaceMethod("[[", signature(x = "SoilProfileCollection"),
                  })
 
 
-#' Get data from column of horizon or site data in a SoilProfileCollection
+#' @title Get data from column of horizon or site data in a SoilProfileCollection
+#'
 #' @name $
-#' @description Get the data from a column accessed by name \code{x$name}. Column names other than profile ID are not shared between site and horizons.
+#' @description  Get the data from a column accessed by name `x$name`. Column
+#'   names other than profile ID are not shared between site and horizons.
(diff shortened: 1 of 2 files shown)


dylanbeaudette:
Nice detective work, much more informative auto-complete.

